PR #12, opened 2021-06-22
Fix the remote trigger of Pipeline

LinuxSuRen:


LinuxSuRen:
I prefer to remove this feature. Because it requires the auth token. In this case. There are no differences between the remote trigger and API calls.

PR #15, opened 2021-06-24
Add s2i dependency into ks-devops chart

JohnNiang:
Signed-off-by: johnniang <johnniang@fastmail.com>

Adding missing s2i dependency into ks-devops chart fixes dependencies error while executing `helm lint`.


LinuxSuRen (approved):
LGTM

PR #25, opened 2021-06-30
Add an all in one command which combines apiserver and controller-manager

LinuxSuRen:


cmd/allinone/app/entrypoint.go
```diff
@@ -0,0 +1,75 @@
...
+
+	// Load configuration from file
+	conf, err := config.TryLoadFromDisk()
+	if err == nil {
+		s = &options.ServerRunOptions{
+			GenericServerRunOptions: s.GenericServerRunOptions,
+			Config:                  conf,
+		}
+	} else {
+		klog.Fatal("Failed to load configuration from disk", err)
```
JohnNiang, line 43: It's better to log as warning or break current thread.

cmd/allinone/app/entrypoint.go
```diff
@@ -0,0 +1,75 @@
...
+		s = &controllerOpt.DevOpsControllerManagerOptions{
+			KubernetesOptions: conf.KubernetesOptions,
+			JenkinsOptions:    conf.DevopsOptions,
+			S3Options:         conf.S3Options,
+			LeaderElection:    s.LeaderElection,
+			LeaderElect:       s.LeaderElect,
+			WebhookCertDir:    s.WebhookCertDir,
+		}
+	} else {
+		klog.Fatal("Failed to load configuration from disk", err)
```
JohnNiang, line 68: See above.

cmd/allinone/main.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"kubesphere.io/devops/cmd/allinone/app"
+	"log"
+)
+
+func main() {
+	cmd := app.NewCommand()
+	if err := cmd.Execute(); err != nil {
+		log.Fatalln(err)
```
JohnNiang, line 11: Need append `os.Exit(1)` below here.

JohnNiang (changes requested):

cmd/allinone/app/entrypoint.go
LinuxSuRen: Actually, it's not true. It cannot continue without a configuration.

cmd/allinone/main.go
LinuxSuRen: It's ok if it returns an error code once the command line exit with an error.

image

PR #27, opened 2021-07-01
Add oauth/authenticate API for Jenkins plugin token auth

LinuxSuRen:
I copied the API `oauth/autenticate` from [kubesphere](https://github.com/kubesphere/kubesphere/tree/master/pkg/kapis/oauth).


pkg/constants/constants.go
```diff
@@ -129,7 +129,7 @@ const (
 	NotificationTag             = "Notification"
 	NotificationSecretNamespace = "kubesphere-monitoring-federated"
 	NotificationManagedLabel    = "notification.kubesphere.io/managed"
-	K8SToken = "ks8.token"
+	K8SToken                    = "ks8.token"
```
JohnNiang, line 132: ks8 -> k8s.

JohnNiang (changes requested):

pkg/constants/constants.go
LinuxSuRen: Done

PR #29, opened 2021-07-05
Replace ksinformer with ksclient in devops operator implementation

JohnNiang:
Due to change of devops authentication, the informer can not access resources correctly. Therefore, we have to replace ksinformer with ksclient to make our apis work normally.


pkg/models/devops/devops.go
```diff
@@ -190,7 +187,7 @@ func (d devopsOperator) CreateDevOpsProject(workspace string, project *v1alpha3.
 }
 
 func (d devopsOperator) GetDevOpsProject(workspace string, projectName string) (*v1alpha3.DevOpsProject, error) {
-	return d.ksInformers.Devops().V1alpha3().DevOpsProjects().Lister().Get(projectName)
+	return d.ksclient.DevopsV1alpha3().DevOpsProjects().Get(context.Background(), projectName, metav1.GetOptions{})
```
LinuxSuRen, line 190: `devopsOperator` already has a `context`. So, please use `d.context` instead of creating a new one `context.Background()`

pkg/models/devops/devops.go
```diff
@@ -207,14 +204,14 @@ func (d devopsOperator) UpdateDevOpsProject(workspace string, project *v1alpha3.
 }
 
 func (d devopsOperator) ListDevOpsProject(workspace string, limit, offset int) (api.ListResult, error) {
-	data, err := d.ksInformers.Devops().V1alpha3().DevOpsProjects().Lister().List(labels.Everything())
+	devOpsProjectList, err := d.ksclient.DevopsV1alpha3().DevOpsProjects().List(context.Background(), metav1.ListOptions{})
```
LinuxSuRen, line 207: Same with above.

LinuxSuRen (changes requested):
Consider you already removed it. So it should not be here any more.

image

pkg/models/devops/devops.go
JohnNiang: Make sense to me.

pkg/models/devops/devops.go
JohnNiang: Alright.

JohnNiang:
> Consider you already removed it. So it should not be here any more.
> 
> image

I've considered this removal, but I have to reconfirm that we will use it again in the future.

pkg/models/devops/devops.go
JohnNiang: Actually, it's not true. Function `context#Background()` always returns an initialized, non-nil, empty Context.

Source code of context#Background() as follow:

```go
var (
	background = new(emptyCtx)
	todo       = new(emptyCtx)
)

// Background returns a non-nil, empty Context. It is never canceled, has no
// values, and has no deadline. It is typically used by the main function,
// initialization, and tests, and as the top-level Context for incoming
// requests.
func Background() Context {
	return background
}
```

LinuxSuRen: Why do you always need an initialized, non-nil, empty Context?

LinuxSuRen: By the way, you now can rebase with the master branch. I fixed the unit test failure from #28

PR #69, opened 2021-07-22
Change the default registry to multiple registries (#61)

### LINKED ISSUE #61: Change default docker registry into docker.io for charts
Currently, the default docker registry of ks-devops charts is `ghcr.io`, which is too slow if you are in Chinese network.

See also: https://github.com/kubesphere-sigs/ks-devops/blob/678147e3ae80f3f12370ee743c03d2041f212864/charts/ks-devops/values.yaml

### Suggestions

Change the default docker registry from `ghcr.io` into `docker.io`.

/good-first-issue

shihaoH:
Fix #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ Want to go into deep? Please checkout the [documentation](docs).
 ### Install it as a Helm Chart
 
 First, please clone this git repository. Then run command: `make install-chart`
+ - Because the default registry uses `ghcr.io/kubesphere-sigs`, if you want to use `docker.io` as the registry for image pull, you can exec `helm install ks-devops chart/ks-devops --set image.registry=kubespheresig`
 
 ### Run it locally
 
--- a/charts/ks-devops/values.yaml
+++ b/charts/ks-devops/values.yaml
@@ -1,14 +1,16 @@
 replicaCount: 1
 
 image:
+  # support 'ghcr.io/kubesphere-sigs' or 'kubespheresig'
+  registry: kubespheresig
   controller:
-    repository: ghcr.io/kubesphere-sigs/devops-controller
+    repository: devops-controller
     tag: master
   apiserver:
-    repository: ghcr.io/kubesphere-sigs/devops-apiserver
+    repository: devops-apiserver
     tag: master
   tools:
-    repository: ghcr.io/kubesphere-sigs/devops-tools
+    repository: devops-tools
     tag: master
   pullPolicy: IfNotPresent
 
--- a/charts/ks-devops/values.yaml
+++ b/charts/ks-devops/values.yaml
@@ -2,7 +2,7 @@ replicaCount: 1
 
 image:
   # support 'ghcr.io/kubesphere-sigs' or 'kubespheresig'
-  registry: kubespheresig
+  registry: ghcr.io/kubesphere-sigs
   controller:
     repository: devops-controller
     tag: master
--- a/charts/ks-devops/templates/deployment-apiserver.yaml
+++ b/charts/ks-devops/templates/deployment-apiserver.yaml
@@ -34,7 +34,7 @@ spec:
         {{- toYaml .Values.podSecurityContext | nindent 8 }}
       initContainers:
         - name: "init-config"
-          image: "{{ .Values.image.tools.repository }}:{{ .Values.image.tools.tag | default .Chart.AppVersion }}"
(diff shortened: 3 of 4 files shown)


charts/ks-devops/values.yaml
```diff
@@ -1,14 +1,16 @@
 replicaCount: 1
 
 image:
+  # support 'ghcr.io/kubesphere-sigs' or 'kubespheresig'
+  registry: kubespheresig
```
LinuxSuRen, line 5: As we discussed in https://github.com/kubesphere-sigs/ks-devops/issues/61#issuecomment-882937563

I prefer to use `ghcr.io` as the default registry.

Normally, people read the usage from [README file](https://github.com/kubesphere-sigs/ks-devops/tree/master/charts/ks-devops). So, please add a description of this value into it.

LinuxSuRen (changes requested):
Thanks for your quick response. I left some comments.

And please see the following screenshot. I guess you didn't use your email which connected with your GitHub account. So it's a gray icon.

image

JohnNiang:
Hi @LinuxSuRen , please re-run GitHub Checks, because the lint stage was time-out. I will increase timeout later.

charts/ks-devops/values.yaml
shihaoH: OK, I will add a description later

README.md
```diff
@@ -18,6 +18,7 @@ Want to go into deep? Please checkout the [documentation](docs).
 ### Install it as a Helm Chart
 
 First, please clone this git repository. Then run command: `make install-chart`
+ - Because the default registry uses `ghcr.io/kubesphere-sigs`, if you want to use `docker.io` as the registry for image pull, you can modify `chart/ks-devops/values.yaml image.registry`
```
LinuxSuRen, line 21: It's incorrect. People don't need to modify anything. They just need to pass parameters into the helm command. For example:

`helm install ks-devops . --set image.registry=docker.io`

shihaoH: or how to add parameters in `make install-chart`

LinuxSuRen: There're many parameters in the helm chart. Normally, all the descriptions should be able to find in the readme file of the helm chart.

image

And in most use cases, people install helm charts by the helm command instead of the make command. So I don't recommend do more things on the make command. The make command is only useful for the developers.

shihaoH: OK

LinuxSuRen (commented):
/lgtm

LinuxSuRen (commented):
/lgtm

README.md
```diff
@@ -18,6 +18,7 @@ Want to go into deep? Please checkout the [documentation](docs).
 ### Install it as a Helm Chart
 
 First, please clone this git repository. Then run command: `make install-chart`
+ - Because the default registry uses `ghcr.io/kubesphere-sigs`, if you want to use `docker.io` as the registry for image pull, you can exec `helm install ks-devops chart/ks-devops --set image.registry=kubespheresig`
```
JohnNiang, line 21: In fact, `ghcr.io` is registry address instead of `ghcr.io/kubesphere-sigs`.

JohnNiang (changes requested):

README.md
LinuxSuRen: Good catch. But the name path of dockerhub doesn't allow `-`. So there's no way to adapt `docker.io/kubesphere-sigs/xxx`. How about the following naming:

```
images:
  # support 'ghcr.io/kubesphere-sigs' or 'kubespheresig'
  repository: ghcr.io/kubesphere-sigs
  controller:
    image: devops-controller
    tag: master
```

AbdelouahabMbarki (approved):
/lgtm

README.md
AbdelouahabMbarki: I can pull images using  links like that docker.io/kubesphere-sigs/xxx  with docker  but i don't know if that's the case for kubernetes too

AbdelouahabMbarki: This solution looks good

JohnNiang: @AbdelouahabMbarki please see here: https://docs.docker.com/docker-id/

AbdelouahabMbarki: I see , thanks .

LinuxSuRen: If you think we should change the naming? @JohnNiang

JohnNiang: It would be better to keep the status quo.

PR #71, opened 2021-07-22
Extract authentication jwtSecret as a variable

### LINKED ISSUE #70: Make authentication.jwtSecret configurable
Recently, I was trying to integrate ks-devops and ks-devops-plugin into ks-installer, but something unexpected happened. When I went into devops projects page, I was kicked off the line.

At last, I found that authentication.jwtSecret was hard coded as below:

- ks-devops

https://github.com/kubesphere-sigs/ks-devops/blob/8e47133c2250165faeb07d5f12ce1873680c3ec9/charts/ks-devops/templates/config.yaml#L9

- [ks-devops-plugin](https://github.com/kubesphere-sigs/ks-devops-plugin/blob/297d271f98acd7ff7546e41b9e47461f375dd3ea/charts/ks-devops-plugin/templates/config.yaml#L9)

So we should make them configurable out of helm template to adapt other third systems.

JohnNiang:
### Why we need it

If we hard code authentication jwtSecret in helm charts, the jwt tools in init container of ks-devops-apiserver won't update jwt secret correctly, at least we should set the jwt secret with empty. If we integrate with KubeSphere, consistent secrets are required among ks-devops, ks-devops-plugin and KubeSphere, there is no way to set secret in ks-devops and ks-devops-plugin same as KubeSphere.

### What this PR does:

Extract authentication jwtSecret as a variable. Please refer to #70 for more information.

Fix #70 

/kind bug

### DIFF
--- a/charts/ks-devops/values.yaml
+++ b/charts/ks-devops/values.yaml
@@ -13,7 +13,7 @@ image:
   pullPolicy: IfNotPresent
 
 authentication:
-  # Leave
+  # If not set, jwt tools will generate and set it automatically
   jwtSecret: ""
 
 imagePullSecrets: []
--- a/charts/ks-devops/values.yaml
+++ b/charts/ks-devops/values.yaml
@@ -12,6 +12,10 @@ image:
     tag: master
   pullPolicy: IfNotPresent
 
+authentication:
+  # Leave
+  jwtSecret: ""
+
 imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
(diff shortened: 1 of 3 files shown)


charts/ks-devops/values.yaml
```diff
@@ -12,6 +12,10 @@ image:
     tag: master
   pullPolicy: IfNotPresent
 
+authentication:
+  # Leave
```
LinuxSuRen, line 16: `Leave` ? Missing somgthing?

LinuxSuRen (changes requested):

charts/ks-devops/values.yaml
JohnNiang: My bad, I will refine the comment.

LinuxSuRen (commented):
/lgtm

AbdelouahabMbarki (approved):
/lgtm

PR #72, opened 2021-07-23
WIP: Make kubesphere authentication server configurable

JohnNiang:
### Why we need it

The out of dated ksAuth.server config item is configured in Jenkins charts due to apiserver full name changed of ks-devops. This behaviour will lead to out of contact between KubeSphere and Jenkins. 

### What this PR dose

- Change helm chart name of ks-devops from `ks-ctl` into `ks-devops`
- Make KubeSphere Authentication server configured in Jenkins chart as variable, which could be configured from outside

### What kind of this PR

/kind bug

### DIFF
--- a/charts/ks-devops/charts/jenkins/values.yaml
+++ b/charts/ks-devops/charts/jenkins/values.yaml
@@ -453,4 +453,4 @@ securityRealm:
 
 ksAuth:
   enabled: false
-#  server:
+  server: ""
--- a/Makefile
+++ b/Makefile
@@ -48,8 +48,8 @@ lint-chart:
 install-chart: lint-chart
 	helm install ks-devops charts/ks-devops -n kubesphere-devops-system --set serviceAccount.create=true --create-namespace \
 		--set image.pullPolicy=Always \
-		--set jenkins.ksAuth.enabled=true \
-		--set jenkins.ksAuth.server=http://ks-devops-apiserver.kubesphere-devops-system:9090
+		--set jenkins.ksAuth.enabled=true
+
 uninstall-chart:
 	make uninstall-jenkins-chart || true
 	helm uninstall ks-devops -n kubesphere-devops-system
--- a/Makefile
+++ b/Makefile
@@ -46,14 +46,15 @@ lint-chart:
 	helm lint charts/ks-devops
 
 install-chart: lint-chart
-	helm install ks-ctl charts/ks-devops -n kubesphere-devops-system --set serviceAccount.create=true --create-namespace \
+	helm install ks-devops charts/ks-devops -n kubesphere-devops-system --set serviceAccount.create=true --create-namespace \
 		--set image.pullPolicy=Always \
-		--set jenkins.ksAuth.enabled=true
+		--set jenkins.ksAuth.enabled=true \
+		--set jenkins.ksAuth.server=http://ks-devops-apiserver.kubesphere-devops-system:9090
 uninstall-chart:
 	make uninstall-jenkins-chart || true
-	helm uninstall ks-ctl -n kubesphere-devops-system
+	helm uninstall ks-devops -n kubesphere-devops-system
 render-chart:
-	helm template ks-ctl charts/ks-devops  -n kubesphere-devops-system \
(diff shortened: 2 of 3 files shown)


charts/ks-devops/charts/jenkins/values.yaml
```diff
@@ -453,4 +453,4 @@ securityRealm:
 
 ksAuth:
   enabled: false
-#  server:
+  server: ""
```
LinuxSuRen, line 456: How users can know what the server is?

LinuxSuRen (changes requested):
You do delete the potential incorrect service name. But it will be empty if users do not pass any values for it. And there's no any documents about what the server value should be.

charts/ks-devops/charts/jenkins/values.yaml
JohnNiang: Users could check server from [here](https://github.com/kubesphere-sigs/ks-devops/blob/1102edae144c1dbe4c2005059be9e3b7daaa2bb2/charts/ks-devops/templates/service-apiserver.yaml):

LinuxSuRen: > Users could check server from [here](https://github.com/kubesphere-sigs/ks-devops/blob/1102edae144c1dbe4c2005059be9e3b7daaa2bb2/charts/ks-devops/templates/service-apiserver.yaml).

Why do they should know it? Is there any document for it?

The most important thing is that it should work well without any extra parameters. We must make it be simple as possible as we can.

JohnNiang: We can pass the ksAuth.server outside of charts as below:

```bash
helm install ks-devops -n kubesphere-devops-system xyz
    --set jenkins.ksAuth.enabled=true \
    --set jenkins.ksAuth.server=http://ks-devops-apiserver.kubesphere-devops-system:9090
```

LinuxSuRen: Let's assume I don't have too much knowledge about helm charts and k8s. Then I don't know how to provide `http://ks-devops-apiserver.kubesphere-devops-system:9090`.

JohnNiang: Now, `ksAuth.eanble` config is set with `false` in `Makefile`, and there is no need to concern `ksAuth.server` config when someone runs the command `make install-chart`.

LinuxSuRen: Please answer my question:

Let's assume I don't have too much knowledge about helm charts and k8s. Then I don't know how to provide http://ks-devops-apiserver.kubesphere-devops-system:9090.

JohnNiang: I'will add another documentation to describe how to develop  in KubeSphere in the future.

LinuxSuRen: > I'will add another documentation to describe how to develop in KubeSphere in the future.

I cannot accept it. Users don't need to know about it. Because it's possible to provide a default value. Please see the format:

`http://{ks-devops-chartname}-apiserver.{ks-devops-namespace}:9090`

JohnNiang:
> A subchart is considered "stand-alone", which means a subchart can never explicitly depend on its parent chart.

Please see <https://helm.sh/docs/chart_template_guide/subcharts_and_globals/>.

Consider that users install Jenkins chart only, the rendering result will be unexpected, please see the following:

```bash
# try to render Jenkins chart templates
helm template ks-jenkins ./charts/ks-devops/charts/jenkins | grep ks-devops-apiserver
```

```yaml
# template
server: "http://{{ .Release.Name }}-ks-devops-apiserver.{{ .Release.Namespace }}:9090/"

# rendering result
server: "http://ks-jenkins-ks-devops-apiserver.default:9090/"
```

LinuxSuRen:
> So I suggest that it is necessary to provide a variable `ksAuth.server` here.

it's true.

But the biggest problem is that no one knows what the value `ksAuth.server` should be.

JohnNiang:
Due to https://github.com/kubesphere/ks-devops/pull/176 got merged, this PR can be closed.

/close

PR #74, opened 2021-07-25
Extract agent builder and s2i image's docker registry

JohnNiang:
### Why we need it

Extracting docker registry as one configuration item lets users configure docker image of ks-devops chart easily. Think about if users want to host a private docker registry for docker images related ks-devops.

### What this PR done

- Extract agent builder docker registry
- Extract s2i builder docker registry

/kind improvment

### DIFF
--- a/charts/ks-devops/charts/jenkins/README.md
+++ b/charts/ks-devops/charts/jenkins/README.md
@@ -84,17 +84,17 @@ The following tables list the configurable parameters of the Jenkins chart and t
 ### Jenkins Agent
 
 | Parameter                   | Description                                                                                        | Default                                                                      |
-| --------------------------- | -------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
-| `Agent.AlwaysPullImage`     | Always pull agent container image before build                                                     | `false`                                                                      |
-| `Agent.CustomJenkinsLabels` | Append Jenkins labels to the agent                                                                 | `{}`                                                                         |
-| `Agent.Enabled`             | Enable Kubernetes plugin jnlp-agent podTemplate                                                    | `true`                                                                       |
-| `Agent.Image`               | Agent image name                                                                                   | `jenkinsci/jnlp-slave`                                                       |
-| `Agent.ImagePullSecret`     | Agent image pull secret                                                                            | Not set                                                                      |
-| `Agent.ImageTag`            | Agent image tag                                                                                    | `2.62`                                                                       |
-| `Agent.Privileged`          | Agent privileged container                                                                         | `false`                                                                      |
-| `Agent.resources`           | Resources allocation (Requests and Limits)                                                         | `{requests: {cpu: 200m, memory: 256Mi}, limits: {cpu: 200m, memory: 256Mi}}` |
-| `Agent.volumes`             | Additional volumes                                                                                 | `nil`                                                                        |
-| `Agent.Builder.Registry`    | Agent builder image registry with namespace. Like `ghcr.io/kubesphere` or `docker.io/kubesphere` | `docker.io/kubespheredev`
+| --------------------------- | -------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------- |
+| `Agent.AlwaysPullImage`     | Always pull agent container image before build                                         | `false`                                                                      |
+| `Agent.CustomJenkinsLabels` | Append Jenkins labels to the agent                                                     | `{}`                                                                         |
+| `Agent.Enabled`             | Enable Kubernetes plugin jnlp-agent podTemplate                                        | `true`                                                                       |
+| `Agent.Image`               | Agent image name                                                                       | `jenkinsci/jnlp-slave`                                                       |
+| `Agent.ImagePullSecret`     | Agent image pull secret                                                                | Not set                                                                      |
+| `Agent.ImageTag`            | Agent image tag                                                                        | `2.62`                                                                       |
+| `Agent.Privileged`          | Agent privileged container                                                             | `false`                                                                      |
+| `Agent.resources`           | Resources allocation (Requests and Limits)                                             | `{requests: {cpu: 200m, memory: 256Mi}, limits: {cpu: 200m, memory: 256Mi}}` |
+| `Agent.volumes`             | Additional volumes                                                                     | `nil`                                                                        |
+| `Agent.Builder.Registry`    | Agent builder image registry with namespace. Available values: `docker.io/kubesphere`, `docker.io/kubespheredev`  | `docker.io/kubesphere`                            |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`.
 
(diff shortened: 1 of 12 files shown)


charts/ks-devops/charts/jenkins/README.md
```diff
@@ -83,17 +83,18 @@ The following tables list the configurable parameters of the Jenkins chart and t
...
+| `Agent.ImageTag`            | Agent image tag                                                                                    | `2.62`                                                                       |
+| `Agent.Privileged`          | Agent privileged container                                                                         | `false`                                                                      |
+| `Agent.resources`           | Resources allocation (Requests and Limits)                                                         | `{requests: {cpu: 200m, memory: 256Mi}, limits: {cpu: 200m, memory: 256Mi}}` |
+| `Agent.volumes`             | Additional volumes                                                                                 | `nil`                                                                        |
+| `Agent.Builder.Registry`    | Agent builder image registry with namespace. Like `ghcr.io/kubesphere` or `[docker.io/]kubesphere` | `kubespheredev`
```
LinuxSuRen, line 97: I know the prefix `docker.io` is optional if the container runtime is Docker. But I prefer to use `docker.io/kubesphere`. There're two pros:

* There's no potential confusion for anyone
  * Some users might don't understand the meaning of `[docker.io/]kubesphere`
* Other container runtime does not take `docker.io` as the default prefix, such as containerd

LinuxSuRen (commented):
I left some suggestions. Please take a look at it.

charts/ks-devops/charts/jenkins/README.md
JohnNiang: Thanks for you suggestion. I will take the advice.

charts/ks-devops/charts/jenkins/README.md
```diff
@@ -94,7 +94,7 @@ The following tables list the configurable parameters of the Jenkins chart and t
 | `Agent.Privileged`          | Agent privileged container                                                                         | `false`                                                                      |
 | `Agent.resources`           | Resources allocation (Requests and Limits)                                                         | `{requests: {cpu: 200m, memory: 256Mi}, limits: {cpu: 200m, memory: 256Mi}}` |
 | `Agent.volumes`             | Additional volumes                                                                                 | `nil`                                                                        |
-| `Agent.Builder.Registry`    | Agent builder image registry with namespace. Like `ghcr.io/kubesphere` or `[docker.io/]kubesphere` | `kubespheredev`
+| `Agent.Builder.Registry`    | Agent builder image registry with namespace. Like `ghcr.io/kubesphere` or `docker.io/kubesphere` | `docker.io/kubespheredev`
```
LinuxSuRen, line 97: How about my suggestion about the word like? IMO it's not accurate.

charts/ks-devops/charts/s2i/values.yaml
```diff
@@ -2,9 +2,9 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 
-docker:
-  registry: ""
-  group: "kubespheredev"
+image:
+  # docker.io/kubespheredev, ghcr.io/kubespheredev or any other registries
```
LinuxSuRen, line 6: > or any other registries

Really? IMO it's not accurate. How about `gcr.io`?

charts/ks-devops/charts/s2i/templates/_helpers.tpl
```diff
@@ -41,15 +41,3 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
...
-Concrete prefix of s2i image
-*/}}
-{{- define "s2i.imagePrefix" -}}
-{{- if .Values.docker.registry }}
-{{- print .Values.docker.registry "/" .Values.docker.group }}
-{{- else }}
-{{- print .Values.docker.group }}
-{{- end }}
-{{- end }}
-
```
LinuxSuRen, line 55: Why delete these lines?

LinuxSuRen (changes requested):
Please separate this PR into two. Or please update the PR title. Please keep doing one thing in one PR.

charts/ks-devops/charts/jenkins/README.md
JohnNiang: IMO, they are examples for users. Think about if users have hosted a private registry for ks-devops. How about I change to the following?

```
Agent builder image registry with namespace. Example: {docker_registry}/{docker_id}. Available values: `ghcr.io/kubesphere` or `docker.io/kubesphere`.
```

JohnNiang: Another question here: Should we use `kubesphere` or `kubespheredev` as the docker ID here?

charts/ks-devops/charts/s2i/templates/_helpers.tpl
JohnNiang: For simplicity and consistency, I've removed the `registry.group` option here and used `image.registry` as s2i image prefix :

```diff
- docker:
+ image
-   registry: ""
+   registry: "docker.io/kubespheredev"
-   group: "kubespheredev"
```

charts/ks-devops/charts/jenkins/README.md
LinuxSuRen: > Another question here: Should we use `kubesphere` or `kubespheredev` as the docker ID here?

I prefer to don't change too many things here. You just want to support multiple registries here. So please just keep doing one thing.

`{docker_registry}/{docker_id}` is incorrect.  Docker is container runtime instead of equal to container runtime. There are many other container runtimes, for example: `containerd`.

I know you want to provide an example. But the following values is not an example but a real and available value:
* `ghcr.io/kubesphere`

charts/ks-devops/charts/s2i/templates/_helpers.tpl
LinuxSuRen: fair enough

JohnNiang:
hi @LinuxSuRen , I've updated pull request title and description. Meanwhile, I've responded to your comments, please have a look at it..

charts/ks-devops/charts/s2i/values.yaml
LinuxSuRen: @JohnNiang How about this one?

JohnNiang: `gcr.io` is not supported currently. However, my original intention is private registry. I will change it according README.md.

charts/ks-devops/charts/jenkins/README.md
JohnNiang: Make sense to me.

charts/ks-devops/charts/s2i/values.yaml
LinuxSuRen: I can guess what do you want to address here. But we need to make it be accurate.

JohnNiang: Alright.

LinuxSuRen (commented):
/lgtm

PR #89, opened 2021-08-01
Add prometheus service monitor and rule

JohnNiang:
### Why we need it

This PR aims to make users use ks-devops metrics out of box or integrate with Prometheus easily.

### What this PR dose

- Add Prometheus ServiceMonitor and PrometheusRule in Jenkins helm chart
- Add Prometheus ServiceMonitor in S2I helm chart

/kind feature


LinuxSuRen (commented):
/lgtm

PR #99, opened 2021-08-04
Add license file which is the apache 2.0

LinuxSuRen:
Most of the code lines come from https://github.com/kubesphere/kubesphere/. So it's good to keep to use the same license.


LinuxSuRen:
I'll merge this PR if there's no negative feedback on this.

PR #115, opened 2021-08-06
Add a root API path which without the group name

### LINKED ISSUE #116: If kubeconfig certificate and docker user certificate are created, the response is abnormal
image

image

LinuxSuRen:
According to the https://github.com/kubesphere/kubesphere/pull/4129. The new DevOps proxy does not send the HTTP request with the group name of DevOps APIs. So in this PR, I let it support two styles that contain or not the group name.

## How to test it manually

You only need to update the image `surenpi/devops-apiserver:dev--71e1645`. Because this PR didn't change the controller part.

fix #116

### DIFF
--- a/pkg/kapis/devops/v1alpha3/handler.go
+++ b/pkg/kapis/devops/v1alpha3/handler.go
@@ -415,15 +415,20 @@ func (h *devopsHandler) DeleteCredential(request *restful.Request, response *res
 	}
 }
 
-func (h *devopsHandler) getDevOps(request *restful.Request) (devops.DevopsOperator, error) {
+func (h *devopsHandler) getDevOps(request *restful.Request) (operator devops.DevopsOperator, err error) {
 	ctx := request.Request.Context()
 	token := ctx.Value(constants.K8SToken).(string)
 
-	klog.V(9).Infof("get DevOps client with  token: %s", token)
-	kubernetesClient, err := k8s.NewKubernetesClientWithToken(token, h.k8sClient.Config().Host)
-	if err != nil {
-		return nil, err
+	var k8sClient k8s.Client
+	if token == "" {
+		k8sClient = h.k8sClient
+	} else {
+		klog.V(9).Infof("get DevOps client with token: %s", token)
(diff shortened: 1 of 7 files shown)


LinuxSuRen:
ping @kubesphere/sig-devops

JohnNiang:
> ping @kubesphere/sig-devops

I'm testing this PR.

pkg/kapis/devops/v1alpha3/handler.go
```diff
@@ -415,15 +415,20 @@ func (h *devopsHandler) DeleteCredential(request *restful.Request, response *res
...
 	ctx := request.Request.Context()
 	token := ctx.Value(constants.K8SToken).(string)
 
-	klog.V(9).Infof("get DevOps client with  token: %s", token)
-	kubernetesClient, err := k8s.NewKubernetesClientWithToken(token, h.k8sClient.Config().Host)
-	if err != nil {
-		return nil, err
+	var k8sClient k8s.Client
+	if token == "" {
+		k8sClient = h.k8sClient
```
JohnNiang, line 424: Please explain why we allow people to access our resources without token.

LinuxSuRen: as you can see from kubesphere/kubesphere#4129

it does not provide the K8S token. And other modules have the same pattern. Such as, notification manager.

JohnNiang:
BTW, we can config the endpoint with path prefix `/kapis/devops.kubesphere.io/` to resolve this problem. Please see [here](https://github.com/kubesphere/kubesphere/blob/418a2a09c7f751cc6f1bcacbf141c6f64e6aa256/pkg/kapis/generic/generic.go#L102-L114).

LinuxSuRen:
> BTW, according to the [kubesphere/kubesphere#4129](https://github.com/kubesphere/kubesphere/pull/4129), we can config the endpoint with path prefix `/kapis/devops.kubesphere.io/` to resolve this problem. Please see [here](https://github.com/kubesphere/kubesphere/blob/418a2a09c7f751cc6f1bcacbf141c6f64e6aa256/pkg/kapis/generic/generic.go#L102-L114).

Yes. But all APIs of this project might have the same prefix. So do we still need it?

PR #124, opened 2021-08-09
Align go version with version defined in go.mod

JohnNiang:
### What this PR dose

Align go version in Test action with version defined in go.mod.

### Why we need it

Keep same go version all the time.


LinuxSuRen (approved):
/lgtm

Issue #67, opened 2021-07-21
Request fully support containerd (runtime, building time)


LinuxSuRen:
## Background
As we know, [Kubernetes can run on serval different container runtimes](https://kubernetes.io/docs/setup/production-environment/container-runtimes/). For example, it could be [containerd](https://containerd.io/), [docker](https://www.docker.com/), etc. But ks-devops can only support docker right now.

Many users start to use other container runtimes. So, I believe that it's worth it for us to support a second container runtime.

## Propose
[containerd](https://containerd.io/) was graduated from CNCF. It's a wonderful project. I recommend we support it as a second option.

## Potential actions
I didn't have too much experience in this area. So I can only provide some possible actions on this feature request.

* Support run ks-devops on containerd (I guess we don't need to do anything about this, but it's necessary to test it)
* Support build images in the Jenkins pipeline
* Provide a guide for users that don't use docker CLI to build images
  * multi-stage images building support
  * multi-arch images building support

## References
* https://containerd.io/
* https://kubesphere.com.cn/forum/d/3811-kubesphere-devops-docker
* https://kubesphere.com.cn/forum/d/3054-dockerkubernetes
* https://www.redhat.com/sysadmin/podman-inside-container

/help

LinuxSuRen:
Considering [Podman](https://github.com/containers/podman) supports a Docker-compatible CLI interface. We will support Podman. In the meantime, a command alias can reduce the migration work for users.

`alias docker='podman'`

So, basically, users do not need to change anything.

LinuxSuRen:
Steps to create KubeSphere via kk:

Create file `config-sample.yaml` with the following content:
```
kind: Cluster
metadata:
  name: sample
spec:
  hosts:
  - {name: node1, address: 192.168.0.8, internalAddress: 192.168.0.8, user: root}
  roleGroups:
    etcd:
    - node1
    master:
    - node1
    worker:
    - node1
  controlPlaneEndpoint:
    domain: lb.kubesphere.local
    address: ""
    port: 6443
  kubernetes:
    version: v1.19.8
    imageRepo: kubesphere
    clusterName: cluster.local
    containerManager: containerd
  network:
    plugin: calico
    kubePodsCIDR: 10.233.64.0/18
    kubeServiceCIDR: 10.233.0.0/18
  registry:
    registryMirrors: []
    insecureRegistries: []
  addons: []
```

[...]

PR #171, opened 2021-08-11
Add podman support in the Jenkins chart

LinuxSuRen:
This is part of https://github.com/kubesphere/ks-devops/issues/67

You can get details from README file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,8 +53,15 @@ uninstall-chart:
 	make uninstall-jenkins-chart || true
 	helm uninstall ks-devops -n kubesphere-devops-system
 render-chart:
+<<<<<<< HEAD
 	helm template ks-devops charts/ks-devops -n kubesphere-devops-system --create-namespace \
 		 --set image.pullPolicy=Always --set jenkins.ksAuth.enabled=true --debug
+=======
+	helm template ks-ctl charts/ks-devops  -n kubesphere-devops-system \
+		--set serviceAccount.create=true --create-namespace \
+		--set image.pullPolicy=Always \
+		--set jenkins.ksAuth.enabled=true --debug
+>>>>>>> 0606682 (Remove the comment statements in the makefile)
 
 install-jenkins-chart:
 	helm install ks-jenkins-test charts/ks-devops/charts/jenkins --set Master.NodePort=
--- a/Makefile
+++ b/Makefile
@@ -53,15 +53,8 @@ uninstall-chart:
 	make uninstall-jenkins-chart || true
 	helm uninstall ks-devops -n kubesphere-devops-system
 render-chart:
-<<<<<<< HEAD
 	helm template ks-devops charts/ks-devops -n kubesphere-devops-system --create-namespace \
 		 --set image.pullPolicy=Always --set jenkins.ksAuth.enabled=true --debug
-=======
-	helm template ks-ctl charts/ks-devops  -n kubesphere-devops-system \
-		--set serviceAccount.create=true --create-namespace \
-		--set image.pullPolicy=Always \
-		--set jenkins.ksAuth.enabled=true --debug
->>>>>>> 0606682 (Remove the comment statements in the makefile)
 
 install-jenkins-chart:
 	helm install ks-jenkins-test charts/ks-devops/charts/jenkins --set Master.NodePort=
--- a/charts/ks-devops/README.md
+++ b/charts/ks-devops/README.md
@@ -1,5 +1,10 @@
 helm install ks-ctl . -n kubesphere-devops-system --create-namespace
 
-# TODO
+## Dockerless
 
-Add Jenkins, S2I into this chart.
+As we know, Kubernetes can run on several [different container runtimes](https://kubernetes.io/docs/setup/production-environment/container-runtimes/). 
+For example, it could be [containerd](https://github.com/containerd/containerd), docker, etc.
+
+Please take the following parameters if you want to install on a Dockerless environment:
+
+`--set jenkins.Agent.Builder.ContainerRuntime=podman`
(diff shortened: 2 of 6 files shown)


LinuxSuRen:
hi @JohnNiang, would you mind helping to review this PR? Thanks.

Makefile
```diff
@@ -56,6 +56,9 @@ render-chart:
 	helm template ks-ctl charts/ks-devops  -n kubesphere-devops-system \
 		--set serviceAccount.create=true --create-namespace \
 		--set image.pullPolicy=Always \
+#		--set jenkins.Agent.Builder.Registry=ghcr.io/linuxsuren \
+#		--set jenkins.Agent.Builder.Maven.Tag=test-podman \
```
JohnNiang, line 60: test-podman -> test.

JohnNiang (changes requested):

Makefile
LinuxSuRen: See also https://github.com/LinuxSuRen/devops-agent/pkgs/container/builder-maven

image

JohnNiang:
It should be better If we can install chart by executing command like this:

```bash
make install-chart engine=podman
```

LinuxSuRen:
> It should be better If we can install chart by executing command like this:
> 
> ```shell
> make install-chart engine=podman
> ```

The make command does not support that.

Makefile
LinuxSuRen: By the way, the line 59 and 60 is just an example for user if they want to change the image name and tag.

JohnNiang:
If I execute `make render-chart`, a error appears:

```bash
/bin/sh: line 7: --set: command not found
make: *** [Makefile:56: render-chart] Error 127
```

LinuxSuRen:
> make render-chart

It works well in the following environment:

macOS 10
```
helm version
version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"dirty", GoVersion:"go1.15.2"}
```

So, what's the version of helm do you use?

Makefile
JohnNiang: I found that we will append `-podman` to the maven tag by default if we choose podman as container runtime.

https://github.com/kubesphere/ks-devops/blob/6024df1d54657f7a3454ba1db83709d2b883fb96/charts/ks-devops/charts/jenkins/templates/_helpers.tpl#L50-L54

https://github.com/kubesphere/ks-devops/blob/6024df1d54657f7a3454ba1db83709d2b883fb96/charts/ks-devops/charts/jenkins/templates/jenkins-casc-config.yml#L108-L111

And I also found [there](https://github.com/LinuxSuRen/devops-agent/pkgs/container/builder-maven) was a image with `test-podman-podman` tag. But I think it is so weird.

JohnNiang:
> > make render-chart
> 
> It works well in the following environment:
> 
> macOS 10
> 
> ```
> helm version
> version.BuildInfo{Version:"v3.3.4", GitCommit:"a61ce5633af99708171414353ed49547cf05013d", GitTreeState:"dirty", GoVersion:"go1.15.2"}
> ```
> 
> So, what version of helm do you use?

image

Makefile
LinuxSuRen: That's not a problem. As I mentioned, line 59 and 60 are just the example usage.

JohnNiang:
Hi @LinuxSuRen , I've installed KubeSphere in containerd. But I'm confused how can I test against it.

LinuxSuRen:
> Hi @LinuxSuRen , I've installed KubeSphere on containerd. But I'm confused how can I test against it.

See https://github.com/kubesphere/ks-devops/pull/171/files#diff-17a4ca7a93821a7c1460483d3a9ca5ea9e82f8c554d17f52f382553d2a2ee3fbR3

JohnNiang:
> > Hi @LinuxSuRen , I've installed KubeSphere on containerd. But I'm confused how can I test against it.
> 
> See https://github.com/kubesphere/ks-devops/pull/171/files#diff-17a4ca7a93821a7c1460483d3a9ca5ea9e82f8c554d17f52f382553d2a2ee3fbR3
> 
> After you installed DevOps with the above document instructions. You can create a Pipeline to build your image. You can use the command podman or docker (there is a command alias docker=podman).

In my practical operation, testing this PR is too complicated.

LinuxSuRen:
> In my practical operation, testing this PR is too complicated.

What are your steps? IMO, you just only need to install DevOps with a particular parameter. Then create a Pipeline to run it.

Or, do you have any good suggestion?

JohnNiang:
> > In my practical operation, testing this PR is too complicated.
> 
> What are your steps? IMO, you just only need to install DevOps with a particular parameter. Then create a Pipeline to run it.
> 
> Or, do you have any good suggestions?

It might be better that we provide a ks-installer docker image for testing.

LinuxSuRen:
> It might be better that we provide a ks-installer docker image for testing.

Actually, you don't really need a KubeSphere UI to test it. It only relies on some Jenkins agent images. So you can just install the Jenkins helm chart to test.

LinuxSuRen:
It's pretty easy to install DevOps via helm chart. See also https://github.com/LinuxSuRen/ks-devops/issues/2

But there are some issues that still exist. All those issues do not come from this PR.

JohnNiang:
hi @LinuxSuRen , please rebase upstream.

LinuxSuRen:
> hi @LinuxSuRen , please rebase upstream. There are some changes related with docker registry of ks-devops-apiserver and ks-devops-controller.

Done.

JohnNiang:
### Jenkinsfile to test

```groovy
pipeline {
  agent {
    node {
      label 'base'
    }

  }
  stages {
    stage('stage-q719p') {
      steps {
        container('base') {
          sh '''podman ps
docker ps
docker info
podman info'''
        }

      }
    }

  }
}
```

### Test result

`docker` command dose not work while `podman` is okay.

image

LinuxSuRen:
The alias command does not work currently. In order to make it work well, we need to build a new image. See also https://github.com/kubesphere/devops-agent/blob/fb80a72c6249ef73344d676509075499a19d5427/base/podman/Dockerfile#L62

No matter what, `podman` is our expected.

JohnNiang:
We should be able to ship.

/lgtm

PR #176, opened 2021-08-14
Fix incorrect variable references and reset important default values

seanly:
1. serviceAccount.create=true: Since ks-devops is a separate service that requires access to kubernetes related resources by default, this variable should be turned on by default
2. jenkins.ksAuth.enabled=false: jenkins.ksAuth.enabled is true must be enabled only when the chart is deployed with ks-Installer. If the chart is used separately, configure --set jenkins.enabled=true --set jenkins.ksAuth.enabled=true jenkins.ksAuth.server=xxx


seanly:
fix #72

LinuxSuRen:
hi @seanly thanks for your contribution! It's great. I will test it later. So it might take some time.

seanly:
> hi @seanly thanks for your contribution! It's great. I will test it later. So it might take some time.

ok

charts/ks-devops/values.yaml
```diff
@@ -44,10 +44,6 @@ securityContext: {}
   # runAsNonRoot: true
   # runAsUser: 1000
 
-service:
```
JohnNiang, line 47: Why should we delete service config?

charts/ks-devops/values.yaml
```diff
@@ -24,7 +24,7 @@ fullnameOverride: ""
 
 serviceAccount:
   # Specifies whether a service account should be created
-  create: false
+  create: true
```
JohnNiang, line 27: IMO, keep this value original is great unless you tell us why we make it true as default.

charts/ks-devops/values.yaml
```diff
@@ -12,7 +12,7 @@ image:
   tools:
     repository: devops-tools
     tag: master
-  pullPolicy: IfNotPresent
+  pullPolicy: Always
```
JohnNiang, line 15: Considering normal users, default pull policy should be `IfNotPresent`.

charts/ks-devops/templates/ingress.yaml
```diff
@@ -10,7 +9,7 @@ kind: Ingress
 metadata:
   name: {{ $fullName }}
   labels:
-    {{- include "ks-devops.labels" . | nindent 4 }}
+    {{- include "ks-devops-apiserver.labels" . | nindent 4 }}
```
JohnNiang, line 12: Why should we change "ks-devops.labels" to "ks-devops-apiserver.labels"?

charts/ks-devops/charts/jenkins/templates/jenkins-casc-config.yml
```diff
@@ -201,7 +201,11 @@ data:
           size: 20
           ttl: 300
         enabled: {{ .Values.ksAuth.enabled }}
-        server: "http://{{ .Release.Name }}-ks-devops-apiserver.{{ .Release.Namespace }}:9090/"
+        {{- if .Values.ksAuth.server }}
+        server: {{ .Values.ksAuth.server }}
+        {{ else }}
+        server: "http://{{ .Release.Name }}-apiserver.{{ .Release.Namespace }}:9090/"
```
JohnNiang, line 207: There still a name conflict of apiserver. Please see [here](https://github.com/kubesphere/ks-devops/blob/6102c5eb1b7b1515266bc6731c86e1a83179720a/charts/ks-devops/templates/_helpers.tpl#L8-L24).

JohnNiang (changes requested):

charts/ks-devops/templates/ingress.yaml
LinuxSuRen: IMO, only the front-end needs the ingress. So the apiserver does not need it.

Makefile
```diff
@@ -46,17 +46,13 @@ lint-chart:
 	helm lint charts/ks-devops
 
 install-chart: lint-chart
-	helm install ks-ctl charts/ks-devops -n kubesphere-devops-system --set serviceAccount.create=true --create-namespace \
-		--set image.pullPolicy=Always \
-		--set jenkins.ksAuth.enabled=true
+	helm install ks-devops charts/ks-devops -n kubesphere-devops-system --create-namespace
```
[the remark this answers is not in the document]
seanly: The default setting is to ensure that the service works properly. However the --set parameter is adjusted for custom installations.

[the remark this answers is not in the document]
seanly: > I suggest we should keep those default value we set before.

The previous default was not to make the system work properly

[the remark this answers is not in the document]
LinuxSuRen: > The previous default was not to make the system work properly

What is the problem with the previous parameter? It would be great if you can post the error output or other log.

charts/ks-devops/templates/ingress.yaml
seanly: > Why should we change "ks-devops.labels" to "ks-devops-apiserver.labels"?

Without this 'ks-devops.labes'

charts/ks-devops/values.yaml
seanly: By default (create: false), serviceaccount is not created and ks-devops controllers cannot be used

charts/ks-devops/charts/jenkins/templates/jenkins-casc-config.yml
seanly: I suggest we create a service with a specific name for Jenkins to use

charts/ks-devops/values.yaml
seanly: Indeed, not good

charts/ks-devops/templates/_helpers.tpl
```diff
@@ -11,16 +11,15 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{- else }}
-{{- $name := default .Chart.Name .Values.nameOverride }}
-{{- if contains $name .Release.Name }}
-{{- .Release.Name | trunc 63 | trimSuffix "-" }}
-{{- else }}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" }}
+{{ .Release.Name | trunc 63 | trimSuffix "-" }}
 {{- end }}
+
+{{- define "ks-devops-apiserver.fullname" -}}
```
JohnNiang, line 17: I think `ks-devops.apiserver-fullname` is more appropriate than `ks-devops-apiserver.fullname`.

charts/ks-devops/templates/_helpers.tpl
```diff
@@ -11,16 +11,15 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{ include "ks-devops.fullname" . }}-apiserver
 {{- end }}
+
+{{- define "ks-devops-controller.fullname" -}}
```
JohnNiang, line 21: Same above.

charts/ks-devops/templates/_helpers.tpl
```diff
@@ -30,6 +29,19 @@ Create chart name and version as used by the chart label.
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
 {{- end }}
 
+{{/*
+Common labels for ks-devops
```
JohnNiang, line 33: Those labels will be added by helm automatically. We don't have to deal with it manually.

JohnNiang (changes requested):

LinuxSuRen:
> /ping @LinuxSuRen

I try my best to test this PR this week. Thanks for your reminder.

LinuxSuRen (changes requested):
Below are my test steps:
* Install chart via `make install-chart`
* Check if the configuration is correct

Part of the CasC YAML file is:
```
      kubespheretokenauthglobalconfiguration:
        cacheConfiguration:
          size: 20
          ttl: 300
        enabled: true
        server: http://ks-ctl-ks-devops-apiserver.kubesphere-devops-system:9090/
```

then, I try to check the address:
```
/ # ping ks-devops-apiserver.kubesphere-devops-system
PING ks-devops-apiserver.kubesphere-devops-system (10.233.4.27): 56 data bytes
64 bytes from 10.233.4.27: seq=0 ttl=64 time=0.042 ms
^C
--- ks-devops-apiserver.kubesphere-devops-system ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max = 0.042/0.042/0.042 ms
/ # ping ks-ctl-ks-devops-apiserver.kubesphere-devops-system
ping: bad address 'ks-ctl-ks-devops-apiserver.kubesphere-devops-system'
/ #
```

LinuxSuRen (commented):
Everything is ok except the LDAP component is missing. Then I figure it out. DevOps still relies on two more components:

* MinIO (used by s2i)
* OpenLDAP (used by Jenkins login)

Currently, those components were managed by ks-installer. See also [here](https://github.com/kubesphere/ks-installer/blob/d5f52fc2ab8ef7a506e5634fef57afde41db2e0c/roles/common/tasks/common-install.yaml#L8).

So, I'm ok with this PR. @seanly , thanks for your time.

/lgtm

PR #180, opened 2021-08-16
Add devcontainer config file

LinuxSuRen:
GitHub codespaces will read config files from .devcontainer

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,12 @@
+{
+	"name": "Existing Dockerfile",
+	"context": "..",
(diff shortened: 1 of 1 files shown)


JohnNiang:
I'm feeling so sad that I have no entrance of [codespace](https://docs.github.com/en/codespaces) now. : P

LinuxSuRen:
> I'm feeling so sad that I have no entrance of [codespace](https://docs.github.com/en/codespaces) now. : P

You can try to apply it.

See also https://github.com/codespaces

I'm going to merge this later if there's no negative feedback.

PR #186, opened 2021-08-16
Add the desgin document of Jenkins config controller

LinuxSuRen:
See the details from the README file.

### DIFF
--- a/controllers/jenkinsconfig/README.md
+++ b/controllers/jenkinsconfig/README.md
@@ -1,0 +1,71 @@
+Jenkins Configuration Controller
+
+## Features
+
+* Reload Jenkins configuration automatically
+* Easily switch Jenkins configuration between pre-defined and custom
+* Smoothly upgrade Jenkins configuration
+* Jenkins configuration management in GitOps way (TODO)
+
+## Background
+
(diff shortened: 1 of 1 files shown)


controllers/jenkinsconfig/README.md
```diff
@@ -0,0 +1,71 @@
...
+data:
+  jenkins.yaml: |
+    xxx: xxx
+  ks-jenkins.yaml: |
+    xxx: xxx
+```
+
+## How-to
+
+Users should only modify the configuration from `ks-jenkins.yaml`, and make sure the annotation has an expected value
```
JohnNiang, line 70: How about renaming `ks-jenkins.yaml` into `custom-jenkins.yaml`?

LinuxSuRen: The name `custom` and the [formula name](https://github.com/kubesphere/ks-devops/pull/186/files#diff-b57917c9723836c1f9ba3e13eca3dd18f24a92f8f6c9909e60423bb77d625e38R43) `custom` will make people feel confused.

And it might not be a customized config, it also might be a pre-defined config.

JohnNiang: Got it.

PR #190, opened 2021-08-17
Set crumb into request header to Jenkins automatically

JohnNiang:
### What this PR dose

Set crumb into request header to Jenkins.

### Why we need it

1. Currently, if DevOps part of ks-console want to request some Jenkins resources, like `/kapis/devops.kubesphere.io/v1alpha2/tojson`, it have to request `/kapis/devops.kubesphere.io/v1alpha2/crumbissuer` API in advance. This design is not friendly for front-end. If do that, front-end will no longer care if any API needs to set up crumb.

2. Some reconcilers, like `PipelineReconciler, PipelineRunReconciler`, might request Jenkins APIs. It might be more useful if we  uniformly set up crumb before every request to Jenkins.

### Special notes

Because method `SetCrumbForConsumer` depends on `Requester#GetJSON` method, it's not easy to write a unit test against the method.

### DIFF
--- a/pkg/client/devops/jenkins/pure_request.go
+++ b/pkg/client/devops/jenkins/pure_request.go
@@ -48,12 +48,22 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 	client := &http.Client{Timeout: 30 * time.Second}
 
 	header := httpParameters.Header
+	if header == nil {
+		header = http.Header{}
+	}
 
 	if j.Requester != nil {
 		auth := j.Requester.BasicAuth
 
 		creds := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", auth.Username, auth.Password)))
 		header.Set("Authorization", fmt.Sprintf("Basic %s", creds))
+
+		// set crumb
+		if err := j.Requester.SetCrumbForConsumer(func(crumbRequestField, crumb string) {
+			header.Set(crumbRequestField, crumb)
+		}); err != nil {
+			return nil, nil, err
+		}
 	}
 	// TODO consider to remove below
 	SetBasicBearTokenHeader(&header)
--- a/pkg/client/devops/jenkins/pure_request.go
+++ b/pkg/client/devops/jenkins/pure_request.go
@@ -47,7 +47,7 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 	apiURL.RawQuery = httpParameters.Url.RawQuery
 	client := &http.Client{Timeout: 30 * time.Second}
 
-	header := httpParameters.Header
+	header := httpParameters.Header.Clone()
 	if header == nil {
 		header = http.Header{}
 	}
@@ -64,7 +64,7 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 			if err := j.Requester.SetCrumbForConsumer(func(crumbRequestField, crumb string) {
 				header.Set(crumbRequestField, crumb)
 			}); err != nil {
-				klog.Errorf("unable to set crumb to http parameters, err: %v", err)
+				klog.Errorf("unable to set crumb to HTTP header, err: %v", err)
 				return nil, nil, err
 			}
 		}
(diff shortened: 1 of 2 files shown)


LinuxSuRen (changes requested):
Only the [POST request](https://github.com/jenkinsci/jenkins/blob/ca646b3742e6921d332078d0b287fd359970d7e9/core/src/main/java/hudson/security/csrf/CrumbFilter.java#L126) requires a crumb.

JohnNiang:
> Only the [POST request](https://github.com/jenkinsci/jenkins/blob/ca646b3742e6921d332078d0b287fd359970d7e9/core/src/main/java/hudson/security/csrf/CrumbFilter.java#L126) requires a crumb.

Hi @LinuxSuRen , thanks for pointing out this potential issue. I have fixed it in latest commit.

pkg/client/devops/jenkins/pure_request.go
```diff
@@ -48,12 +48,26 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
...
 		creds := base64.StdEncoding.EncodeToString([]byte(fmt.Sprintf("%s:%s", auth.Username, auth.Password)))
 		header.Set("Authorization", fmt.Sprintf("Basic %s", creds))
+
+		// we only set crumb to request header for POST method
+		// See also https://github.com/jenkinsci/jenkins/blob/ca646b3742e6921d332078d0b287fd359970d7e9/core/src/main/java/hudson/security/csrf/CrumbFilter.java#L126
+		if httpParameters.Method == http.MethodPost {
+			if err := j.Requester.SetCrumbForConsumer(func(crumbRequestField, crumb string) {
+				header.Set(crumbRequestField, crumb)
+			}); err != nil {
+				klog.Errorf("unable to set crumb to http parameters, err: %v", err)
```
LinuxSuRen, line 67: `http` -> `HTTP`

Crumb should be in the HTTP header instead of parameters.

pkg/client/devops/jenkins/pure_request.go
```diff
@@ -48,12 +48,26 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 	client := &http.Client{Timeout: 30 * time.Second}
 
 	header := httpParameters.Header
+	if header == nil {
+		header = http.Header{}
```
LinuxSuRen, line 52: I don't think this code line can work. After this line, the variable `header` will be a new one.

LinuxSuRen (changes requested):

pkg/client/devops/jenkins/pure_request.go
JohnNiang: If header is nil, it should be a new one(map[string]string).

LinuxSuRen: Yes, but it would update to `httpParameters.Header`

pkg/client/devops/jenkins/pure_request.go
JohnNiang: Alright.

pkg/client/devops/jenkins/pure_request.go
JohnNiang: In fact, the header should be cloned from `httpParameters.Header`. I will fix it later.

pkg/client/devops/jenkins/pure_request.go
```diff
@@ -47,13 +47,27 @@ func (j *Jenkins) SendPureRequestWithHeaderResp(path string, httpParameters *dev
 	apiURL.RawQuery = httpParameters.Url.RawQuery
 	client := &http.Client{Timeout: 30 * time.Second}
 
-	header := httpParameters.Header
+	header := httpParameters.Header.Clone()
```
LinuxSuRen, line 50: It will panic if the header is nil. If the header never be nil, then you don't need to check it `if header == nil {`

JohnNiang: Please note that nil Is a valid method receiver value, see below sample code:

```go
package main

import (
	"fmt"
)

type Header map[string][]string

// Clone returns a copy of h or nil if h is nil.
func (h Header) Clone() Header {
	if h == nil {
		return nil
	}

	// Find total number of values.
	nv := 0
	for _, vv := range h {
		nv += len(vv)
	}
	sv := make([]string, nv) // shared backing array for headers' values
	h2 := make(Header, len(h))
	for k, vv := range h {
		n := copy(sv, vv)
		h2[k] = sv[:n:n]
		sv = sv[n:]
	}
	return h2
}

func main() {
	fmt.Println("Hello, playground")
	var header Header = nil
	header.Clone()
}
```

LinuxSuRen: Learned from it. But I still feel confused.

LinuxSuRen (commented):
/lgtm

PR #193, opened 2021-08-18
fix(cmd/controller cmd/apiserver): remove term which depends on docker library

### LINKED ISSUE #192: Lint failed due to outdated package usage of term
The current workflow of checking code, in the process of a pull request, failed due to outdated package usage.

An error message in my pull request is at [here](https://github.com/kubesphere/ks-devops/actions/runs/1135689078).

The main reason is:
```bash
Lint
SA1019: package github.com/docker/docker/pkg/term is deprecated: use github.com/moby/term instead (staticcheck)
```

By searching in the whole repository, the related source code is [this](https://github.com/kubesphere/ks-devops/blob/481526eff7e6e7c996cd85e1ad5720797b4bd76d/pkg/utils/term/term.go#L23) and the related function [`TerminalSize`](https://github.com/kubesphere/ks-devops/blob/481526eff7e6e7c996cd85e1ad5720797b4bd76d/pkg/utils/term/term.go#L29) is used at [`cmd/controller/server.go`](https://github.com/kubesphere/ks-devops/blob/481526eff7e6e7c996cd85e1ad5720797b4bd76d/cmd/controller/app/server.go#L83) and [`cmd/apiserver/server.go`](https://github.com/kubesphere/ks-devops/blob/481526eff7e6e7c996cd85e1ad5720797b4bd76d/cmd/apiserver/app/server.go#L69).

Do we need to update the package reference link to ` github.com/moby/term` or to delete related codes?

flamywhale:
resolves #192


cmd/controller/app/server.go
```diff
@@ -79,13 +78,6 @@ func NewControllerManagerCommand() *cobra.Command {
 		fs.AddFlagSet(f)
 	}
 
-	usageFmt := "Usage:\n  %s\n"
```
LinuxSuRen, line 82: How about keeping the help print function? And just remove the term part.

flamywhale: I agree. Term package was used to get the width of the running terminal, in order to beautify the help message printed in the terminal. By setting the max width of the print message, the print function can wrap line according the width. But IMO there is no need to wrap the message. More attention should be paid to the devops core code, not to the beautiful printing help message.

LinuxSuRen (approved):

PR #204, opened 2021-08-24
feat(charts/ks-devops): revise the installation of ks-devops for tekton

flamywhale:


charts/ks-devops/README.md
```diff
@@ -24,8 +24,18 @@ Please refer to Helm's [documentation](https://helm.sh/docs/) to get started.
 
 ## Usage
 
-Once Helm is set up properly, you can install the chart with the release name `ks-ctl` using the following command:
+### Jenkins installtion
```
LinuxSuRen, line 27: `installtion` typo

charts/ks-devops/values.yaml
```diff
@@ -88,7 +88,7 @@ affinity: {}
 
 charts:
   jenkins:
-    enabled: true
+    enabled: false
```
LinuxSuRen, line 91: It should be true. Because Jenkins is still the default option.

LinuxSuRen (changes requested):

charts/ks-devops/README.md
flamywhale: Okay, it's a typing error.

charts/ks-devops/values.yaml
flamywhale: Okay, but if I set it as true, when installing tekton, we are supposed to set it to false.

LinuxSuRen: Users can provide two flags for the installation. Set `jenkins` is false, and `tekton` is true. This is not the perfect solution. But we still need to have a default option.

flamywhale:
* There exists a conflict here. 
    * In the directory `charts/ks-devops/crds`, there are two CRDs, which are `devops.kubesphere.io_pipelines.yaml` and `devops.kubesphere.io_pipelineruns.yaml`. 
    * We can only hold one version at the same time, which means that we can hardly provide a way to install either Jenkins or Tekton due to the version conflict of Pipeline related CRDs.
* What if we provide a shell script called `prepare.sh` to generate CRDs and copy them into `crd` directory?

PR #191, opened 2021-08-17
Complete pipelinerun reconciler partially

JohnNiang:
### What this PR dose

1. Watch PipelineRun CR change and trigger Jenkins build API
2. Loop query pipeline run status from Jenkins

### Why we need it

This PR is part of #41 and is second step of the issue.

### Special notes

This PR have some TODO tasks left, which I will complete step by step later.

PR #207, opened 2021-08-27
WIP: Add cronjob to gc PipelineRuns

LinuxSuRen:
upstream https://github.com/kubesphere-sigs/ks/pull/179 and #191 

In order to avoid too many PipelineRuns to let etcd takes high disk pressure. We should clean up the PipelineRuns regularly.

I offered two options to clean it.

* `maxAge`
  * we only clean those PipelineRun that has finished, and it exists longer than `maxAge`
* `maxCount`
  * the max number of each Pipelines' PipelineRun

We take advantage of CronJob to run the command regularly. Also, users can run it manually.

/kind feature

### DIFF
--- a/charts/ks-devops/values.yaml
+++ b/charts/ks-devops/values.yaml
@@ -12,7 +12,24 @@ image:
   tools:
     repository: devops-tools
     tag: master
+  gc:
+    repository: devops-gc
+    tag: master
   pullPolicy: IfNotPresent
+
+gcJobs:
+  # gcJobs.maxAge -- Max age from which `PipelineRun`s will be deleted
+  maxAge: 168h
+  # gcJobs.maxCount -- Max count from which `PipelineRun`'s will be deleted
+  maxCount: 10
+  # gcJobs.schedule -- Cron expression to periodically delete `PipelineRun`s
+  schedule: "0/30 * * * *"
+  # gcJobs.failedJobsHistoryLimit -- Drives the failed jobs history limit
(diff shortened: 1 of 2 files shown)


charts/ks-devops/values.yaml
```diff
@@ -12,8 +12,25 @@ image:
   tools:
     repository: devops-tools
     tag: master
+  gc:
+    repository: devops-gc
+    tag: master
   pullPolicy: IfNotPresent
 
+gcJobs:
```
flamywhale, line 20: Does we need to add descriptions about GC in the `README.md`?

LinuxSuRen: Yes

liangyuanpeng (commented):
I wonder that if the user can see the hint somewhere after the job fails?

LinuxSuRen:
> wonder that if the user can see the hint somewhere after the job fails

Users can see the log output from the failed jobs.

JohnNiang:
Helm charts of DevOps has been moved to https://github.com/kubesphere-sigs/ks-devops-helm-chart. See also https://github.com/kubesphere/ks-devops/pull/229.

LinuxSuRen:
> Helm charts of DevOps has been moved to https://github.com/kubesphere-sigs/ks-devops-helm-chart. See also #229.

Thanks for your reminder. I'm going to close this PR. And replace it with https://github.com/kubesphere-sigs/ks-devops-helm-chart/pull/7

/close

PR #224, opened 2021-09-05
Add Jenkins client in controller

wxharry:
What?
* Get rid of the previous way of loading Jenkins config.
* Add jenkins client to controller and load config just like the other clients.

Why?
* All the clients should load config in the same way. Jenkins client shouldn't read from files directly.
* We should create a Jenkins client at the beginning, but not initiate a client everytime we create a pipeline.

### DIFF
--- a/pkg/client/devops/fake/fakeJenkinsClient.go
+++ b/pkg/client/devops/fake/fakeJenkinsClient.go
@@ -1,0 +1,72 @@
+package fake
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+	"strings"
+
+	"github.com/jenkins-zh/jenkins-client/pkg/core"
+	"github.com/jenkins-zh/jenkins-client/pkg/job"
+	"github.com/jenkins-zh/jenkins-client/pkg/mock/mhttp"
+	"kubesphere.io/devops/pkg/client/devops/jclient"
+)
+
+func NewFakeJenkinsClient(roundTripper *mhttp.MockRoundTripper) (j *jclient.JenkinsClient) {
+	core := core.JenkinsCore{
+		URL:          "http://localhost",
+		UserName:     "",
+		Token:        "",
+		RoundTripper: roundTripper,
+	}
+	j = &jclient.JenkinsClient{
+		Core: core,
+	}
+	return
+}
+
+// PrepareForGetProjectPipeline404 only for test
+func PrepareForGetProjectPipeline404(roundTripper *mhttp.MockRoundTripper, rootURL, user, password string, path, name string) {
+	request, _ := http.NewRequest("GET", fmt.Sprintf("%s/job/%s/job/%s/api/json", rootURL, name, path), nil)
+	response := &http.Response{
+		StatusCode: 404,
+		Proto:      "HTTP/1.1",
+		Request:    request,
+		Body: ioutil.NopCloser(bytes.NewBufferString(`
+		{"name":""}
+		`)),
+	}
+	roundTripper.EXPECT().
+		RoundTrip(core.NewRequestMatcher(request)).Return(response, nil)
+}
+
+// PrepareForCreateProjectPipeline only for test
+func PrepareForCreateProjectPipeline(roundTripper *mhttp.MockRoundTripper, rootURL, user, password string, jobPayload job.CreateJobPayload, path string) {
+	PrepareForGetProjectPipeline404(roundTripper, rootURL, user, password, jobPayload.Name, path)
+
+	playLoadData, _ := json.Marshal(jobPayload)
+	formData := url.Values{
+		"json": {string(playLoadData)},
+		"name": {jobPayload.Name},
+		"mode": {jobPayload.Mode},
+		"from": {jobPayload.From},
+	}
+	payload := strings.NewReader(formData.Encode())
+	path = job.ParseJobPath(path)
+	requestCreate, _ := http.NewRequest(http.MethodPost, fmt.Sprintf("%s/view/all%s/createItem", rootURL, path), payload)
+	requestCreate.Header.Add("Content-Type", "application/x-www-form-urlencoded")
+	requestCreate.Header.Add("CrumbRequestField", "Crumb")
+	responseCreate := &http.Response{
+		StatusCode: 200,
+		Proto:      "HTTP/1.1",
+		Request:    requestCreate,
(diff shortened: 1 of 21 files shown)


JohnNiang:
hi @wxharry , how is it going?

wxharry:
> hi @wxharry , how is it going?

I am still working on the Unit test.

pkg/client/devops/jclient/jenkins.go
```diff
@@ -1,49 +1,28 @@
...
-}
-type Config struct {
-	Jenkins JenkinsConfig `yaml:"devops,omitempty"`
-}
-type JenkinsConfig struct {
-	URL      string `yaml:"host"`
-	UserName string `yaml:"username"`
-	Password string `yaml:"password"`
-	Token    string `yaml:"token,omitempty"`
+	Core core.JenkinsCore
```
JohnNiang, line 9: IMO, `Core core.JenkinsCore` could be replaced by `job.Client` here.

JohnNiang: So that we can use our Jenkins client conveniently instead of creating `job.Client` every invokation.

JohnNiang (changes requested):

pkg/client/devops/jclient/jenkins.go
wxharry: I agree.

JohnNiang: That's my mistake, `job.Client` should be  `core.Client`.

wxharry: we can add job.client here instead of replacing the core. What do you think?

JohnNiang: The `job.Client` is a specific client for a specific function of Jenkins. In jenkins-zh/jenkins-client, there is a [`core.Client`](https://github.com/jenkins-zh/jenkins-client/blob/bc7dc0ac9b3eb3a3ae44ea97dc7f86fffcc33979/pkg/core/core.go#L27-L29) wrapping `core.JenkinsCore`. So we can use it and remove `jclient.JenkinsClient`.

https://github.com/kubesphere/ks-devops/blob/77b5e674169df5edd10bb2d9f167e14de106d0c1/pkg/client/devops/jclient/jenkins.go#L8-L10

wxharry: `jclient.JenkinsClient` also contains functions like `CreateProjectPipeline`, `DeleteProjectPipline`, etc. ([See projectPipeline.go](https://github.com/kubesphere/ks-devops/blob/1cfd8a08c5f16365049395bf9db182a97653715d/pkg/client/devops/jclient/projectPipeline.go)). So, I don't think it can be removed and replaced by core.Client.

pkg/client/devops/jclient/jenkins.go
```diff
@@ -1,49 +1,33 @@
...
-	e = yaml.Unmarshal(yamlFile, &config)
-	if e != nil {
-		return
-	}
-	// capsulate JenkinsCore
-	core = client.JenkinsCore{
-		URL:      config.Jenkins.URL,
-		UserName: config.Jenkins.UserName,
-		Token:    config.Jenkins.Password,
+func NewJenkinsClient(options *jenkins.Options) (jenkinsClient *JenkinsClient, err error) {
```
JohnNiang, line 14: Before this line, I suggest to add code as below to ensure `JenkinsClient` implement `devops.Interface` fully:

```go
var _ devops.Interface = &JenkinsClient{}
```

pkg/client/devops/jclient/role.go
```diff
@@ -0,0 +1,52 @@
+package jclient
+
+import (
+	"kubesphere.io/devops/pkg/client/devops"
+)
+
+// RoleOperator
+
+func (j *JenkinsClient) AddGlobalRole(roleName string, ids devops.GlobalPermissionIds, overwrite bool) error {
+	// To do: Refactor function
```
JohnNiang, line 10: `To do` -> `TODO`

pkg/client/devops/fake/fakeJenkinsClient.go
```diff
@@ -0,0 +1,72 @@
...
+		{"name":""}
+		`)),
+	}
+	roundTripper.EXPECT().
+		RoundTrip(core.NewRequestMatcher(request)).Return(response, nil)
+}
+
+// PrepareForCreateProjectPipeline only for test
+func PrepareForCreateProjectPipeline(roundTripper *mhttp.MockRoundTripper, rootURL, user, password string, jobPayload job.CreateJobPayload, path string) {
+	PrepareForGetProjectPipeline404(roundTripper, rootURL, user, password, jobPayload.Name, path)
```
JohnNiang, line 48: Why invoke this method `PrepareForGetProjectPipeline404` here?

JohnNiang (changes requested):

JohnNiang:
hi @LinuxSuRen , I'm not sure the [test method](https://github.com/kubesphere/ks-devops/pull/224/files#diff-3b7c9f4912aa3ee3f99e2680f751791f65d968ae6b97cad2778a55156d673ba9) is reasonable, please help me review that part.

pkg/client/devops/fake/fakeJenkinsClient.go
wxharry: Because `CreateProjectPipeline` will check whether the job name has been used before creating a job. If not, it will create a job. Method `PrepareForGetProjectPipeline404` gives a 404 response so that `CreateProjectPipeline` will know the job name is available and create a job.

JohnNiang: I'd prefer to invoke `PrepareForGetProjectPipeline404` inside test function as needed.

wxharry: I am sorry. I don't quite understand here. What does `inside test function` mean? Did you mean inside `TestCreateProjectPipeline`?

JohnNiang:
/hold
Hi @wxharry , please execute following commands to merge upstream:

```bash
git fetch upstream
git merge upstream/master
```

And there is a missing method needing to be implemented, please see:

```go
# kubesphere.io/devops/pkg/client/devops/jclient
pkg/client/devops/jclient/jenkins.go:13:5: cannot use &JenkinsClient{} (type *JenkinsClient) as type devops.Interface in assignment:
        *JenkinsClient does not implement devops.Interface (missing ApplyNewSource method)
```

PR #226, opened 2021-09-06
Add config file for codecov

LinuxSuRen:


codecov.yml
```diff
@@ -0,0 +1,4 @@
+# https://docs.codecov.com/docs/ignoring-paths
+ignore:
+  - "*generated*.go"
+  - "**/*generated*.go"         # glob accepted
```
JohnNiang, line 4: Please add `pkg/client/clientset/versioned` into ignoring list.

JohnNiang (changes requested):

PR #227, opened 2021-09-06
Build APIServer, controller manager and tools only when tests passed

JohnNiang:
### What this PR dose

Add `needs` configuration in jobs `BuildController`, `BuildAPIServer` and `BuildTools`.

### Why we need it

Make jobs `BuildController`, `BuildAPIServer` and `BuildTools` buildable only when tests passed. If tests don't pass, the built things are meaningless.

| Before | After |
| --- | --- |
| image|image |

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -23,6 +23,8 @@ on:
 jobs:
   BuildController:
     runs-on: ubuntu-20.04
+    needs:
+      - UnitTest
     steps:
       - uses: actions/checkout@v2
       - name: Docker meta for kubesphere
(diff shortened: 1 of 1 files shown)


LinuxSuRen (changes requested):
Making these tasks be parallel can allow us to get the result as soon as possible. Let's assume, one error bringing into a Dockerfile. The unit tests can pass, but you need to wait more time to get the image build error.

Even worse, what if the unit tests and dockerfile has error, you might need to commit more than twice to know the errors. Becuase the failed unit tests will stop the image build process.

In a short, I don't think we need this.

JohnNiang:
Below are my opinions:

1. In the actual scenario, we will change the test files more frequently than the Dockerfile.
2. If our codes have not beed tested, then building is meaningless. Personally, I think it is a waste of resources.

But it was fair enough. After I saw ArgoCD actions as below, I think we didn't need this as well.

image

/close

PR #231, opened 2021-09-09
Backup git repo to gitee.com

LinuxSuRen:
## Why?
Sometimes, people will meet the network issues of github.com. Or it might be very slow in some areas of China. But it's very fast to visit https://gitee.com/kubesphere/ks-devops for those users.

Secondly, it's possible to bring some users from https://gitee.com as well.

## Tests

See also https://gitee.com/kubesphere/ks-devops/commits/master

and the screenshot:

image

### DIFF
--- a/.github/workflows/backup.yaml
+++ b/.github/workflows/backup.yaml
@@ -1,0 +1,18 @@
+name: Backup
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


JohnNiang:
LGTM

PR #239, opened 2021-09-16
Add action field into PipelineRunSpec

JohnNiang:
### What this PR dose

Add action field into PipelineRunSpec.

### Why we need it

We need to be able to control our PipelineRun.


LinuxSuRen (approved):
/lgtm

PR #242, opened 2021-09-17
Filter nil items in result list

JohnNiang:
### What this PR dose

Filter nil items in result list

### Why we need it

23647d85ed61ee1c9de57cc52c808f1

### Step to test

Docker image for test: `johnniang/devops-apiserver:pipeline-search-bug`

/kind bug


LinuxSuRen (commented):
/lgtm

PR #243, opened 2021-09-17
Add support delete job history records once pipelinerun was deleted

LinuxSuRen:


JohnNiang:
There is a situation we must consider: If we delete the Pipeline, what should we do with the remaining PipelineRuns? Although all PipelineRuns have owener reference with Pipeline, but the Pipeline has been deleted. When PipelineRun finalizers try to invoke Jenkins API, it will fail forever.

controllers/jenkins/pipelinerun/pipelinerun_controller.go
```diff
@@ -186,6 +200,50 @@ func (r *Reconciler) Reconcile(req ctrl.Request) (ctrl.Result, error) {
...
+	if buildNum = getJenkinsBuildNumber(pipelineRun); buildNum < 0 {
+		return
+	}
+
+	jenkinsClient := job.Client{
+		JenkinsCore: r.JenkinsCore,
+	}
+	jobName := fmt.Sprintf("job/%s/job/%s", pipelineRun.Namespace, pipelineRun.Spec.PipelineRef.Name)
+	if err = jenkinsClient.DeleteHistory(jobName, buildNum); err == nil {
+		pipelineRun.ObjectMeta.Finalizers = sliceutil.RemoveString(pipelineRun.ObjectMeta.Finalizers, func(item string) bool {
```
JohnNiang, line 214: IMO, the finalizer removal should be outside this function. This function should keep simplicity and only deletes the external resource.

controllers/jenkins/pipelinerun/pipelinerun_controller.go
```diff
@@ -210,13 +216,12 @@ func (r *Reconciler) deleteJenkinsJobHistory(pipelineRun *prv1alpha3.PipelineRun
...
 	jobName := fmt.Sprintf("job/%s/job/%s", pipelineRun.Namespace, pipelineRun.Spec.PipelineRef.Name)
-	if err = jenkinsClient.DeleteHistory(jobName, buildNum); err == nil {
-		pipelineRun.ObjectMeta.Finalizers = sliceutil.RemoveString(pipelineRun.ObjectMeta.Finalizers, func(item string) bool {
-			return item == prv1alpha3.PipelineRunFinalizerName
-		})
-		err = r.Update(context.TODO(), pipelineRun)
-	} else {
-		err = fmt.Errorf("failed to delete Jenkins job: %s, build: %d, error: %v", jobName, buildNum, err)
+	if err = jenkinsClient.DeleteHistory(jobName, buildNum); err != nil {
+		if strings.Contains(err.Error(), "not found resources") {
```
JohnNiang, line 220: I'd prefer to add a TODO here, we will refactor Jenkins client error which could carry the HTTP response code and original error message.

JohnNiang:
Code looks good from reading it, when I get a chance I'd like to try this out myself locally.

PR #246, opened 2021-09-23
Relocate v1alpha3.PipelineRun type

### LINKED ISSUE #189: Bump the version (v0.2.5) of controller-gen
## Why?
* controller-gen v0.2.5 maintains the CRD file in version `apiVersion: apiextensions.k8s.io/v1beta1`. But other CRD files rely on `apiVersion: apiextensions.k8s.io/v1`. See also [here](https://github.com/kubesphere/ks-devops/blob/63c99e17a2a744c8321a4e71b05889dee151e51b/config/crd/bases/devops.kubesphere.io_pipelines.yaml#L3).
* The latest version of it is v0.6.2
  * v0.2.5 was released on Jan 30, 2020, it's time to consider bumping it

## Suggestions
These files might need to be updated:

* https://github.com/kubesphere/ks-devops/blob/63c99e17a2a744c8321a4e71b05889dee151e51b/Makefile#L157

## TODO
Please consider letting users upgrade controller-gen version.

JohnNiang:
### What type of PR is this?

/kind enhancement

### What this PR dose / Why we need it

- Move v1alpha3.Pipeline type from `pkg/devops/pipelinerun/v1alpha3` into `pkg/devops/v1alpha3`
- Remove unused main.go
- Refactor `Makefile`: Bump version of `controller-gen`
- Re-generate CRDs and `zz_generated.deepcopy.go` by executing command `make controller-gen` and `make generate`

Fixes #189


Makefile
```diff
@@ -116,19 +125,28 @@ mock-gen:
...
-	cd $$CONTROLLER_GEN_TMP_DIR ;\
-	go mod init tmp ;\
-	go get sigs.k8s.io/controller-tools/cmd/controller-gen@v0.6.2 ;\
-	rm -rf $$CONTROLLER_GEN_TMP_DIR ;\
-	}
-CONTROLLER_GEN=$(GOBIN)/controller-gen
-else
-CONTROLLER_GEN=$(shell which controller-gen)
-endif
+CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
```
LinuxSuRen, line 128: why download it to the current directory?

JohnNiang: Ensure that we won't break the existing binary with different verison.

LinuxSuRen: The directory `bin` is temporary. We need to download it again once it was deleted.

JohnNiang: It will automatically download towards bin folder. Please see:

https://github.com/kubesphere/ks-devops/blob/51bd7fbc9ce4afeed32a1a490e8e6193ea6486bb/Makefile#L142-L152

JohnNiang: User can set GOPROXY environment for speeding up download. So I don't think it is a problem.

LinuxSuRen: It's not necessary to download it if contrller-gen existed. And please don't forget `bin` is just a temporary directory. Put it into a temporary is not a mutli-version solution.

JohnNiang: Please see:
 https://github.com/kubesphere/ks-devops/blob/51bd7fbc9ce4afeed32a1a490e8e6193ea6486bb/Makefile#L143

If a binary exists in `bin` folder, the `go-get-tool` won't download it again.

LinuxSuRen: What if I clone this project in different serval directories of my computer, why should I download all these tools many times?
What if I already have `controller-gen` in my PATH, why should I download it again?

JohnNiang: How do we know what's your version of controller-gen or other tools. We have to guarrantee we use controller-gen or other tools with specific version.

BTW,the scenary that users clone repo in different places is unusual.

LinuxSuRen: As I said before `bin` is a temporary directory. So we can delete it at any time. But these tools are not temporary.

About the version, your solution cannot guarantee it too.

JohnNiang: So what's happen if you delete the temporary folder `bin` here? I didn't get your point. It's just like dependencies which we can delete at any time.

LinuxSuRen: I don't want to download these tools everytime when I delete the `bin` directory.

JohnNiang: I'll stick my own view. But for this PR, I will revert the Makefile for stopping controversial discussion.

LinuxSuRen (commented):
/lgtm

Issue #279, opened 2021-09-28
Wrong paging on the Pipeline list page


LinuxSuRen:
![image](https://user-images.githubusercontent.com/1450685/135055927-f21d11ec-ad04-43de-ab83-2e7847eed3bd.png)

This problem comes from https://github.com/kubesphere/ks-devops/blob/625e555ddd5df7f5a2fa5681ebc86af632979b16/pkg/kapis/devops/v1alpha2/devops.go#L77

PR #280, opened 2021-09-28
Fix the wrong using of paging parameter start

LinuxSuRen:
fix https://github.com/kubesphere/ks-devops/issues/279

Container image for the test purpose:
```
surenpi/devops-apiserver:dev-v3.2.0-alpha.0-b60b654
```


pkg/apiserver/query/types_test.go
```diff
@@ -153,3 +180,54 @@ func TestPagination_GetValidPagination(t *testing.T) {
...
+	// valid query selector
+	query := &Query{}
+	assert.Equal(t, labels.NewSelector(), query.Selector())
+
+	// invalid query selector
+	query = &Query{
+		LabelSelector: "a+b",
+	}
+	assert.Equal(t, labels.Everything(), query.Selector())
+}
```
JohnNiang, line 233: New line needed at the end of file.

JohnNiang (changes requested):

pkg/apiserver/query/types_test.go
LinuxSuRen: Done

PR #285, opened 2021-09-29
Enable exported rule of revive for golangci-lint

JohnNiang:
### What this PR dose / why we need it?

- Enable exported rule of revive for golangci-lint

> Please see https://golangci-lint.run/usage/configuration/#config-file

```yaml
issues:
  exclude:
    - EXC0012
  exclude-use-default: false
```

- Exclude some files by hand
- Add some comment on exported fields, functions and methods according to `golangci-lint`.

> But there is a discussion about the `exported` rule, please see
> - https://github.com/golangci/golangci-lint/issues/456.
> - https://github.com/golangci/golangci-lint/issues/2114#issuecomment-878937553

/kind cleanup

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -96,7 +96,54 @@ run:
     - pkg/server/params/params.go
     - pkg/utils/idutils/id_utils.go
     - pkg/utils/idutils/id_utils_test.go
-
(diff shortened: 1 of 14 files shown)


LinuxSuRen (approved):
This is awesome!!! Great!

PR #357, opened 2021-11-12
[Codebase improvement] Reuse handlers devopsProject to devops

Feeeenng:
reuse  handler devopsProject to devops. and queryParam to query

### DIFF
--- a/pkg/kapis/devops/v1alpha3/handler.go
+++ b/pkg/kapis/devops/v1alpha3/handler.go
@@ -192,11 +192,11 @@ func (h *devopsHandler) GetPipeline(request *restful.Request, response *restful.
 }
 
 func (h *devopsHandler) ListPipeline(request *restful.Request, response *restful.Response) {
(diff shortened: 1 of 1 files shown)


JohnNiang:
@Feeeenng , thanks for your first contribution.

/lgtm
/cc @kubesphere/sig-devops

LinuxSuRen (changes requested):
Thanks for your contribution. But please see the following warning (which come from Goland)

image

JohnNiang:
> Thanks for your contribution. But please see the following warning (which come from Goland)
> 
> image

IMO,  some people use `VSCode` as their IDE which lint tool is `golangci-lint`, and no warning here due to [.golangci.yaml](https://github.com/kubesphere/ks-devops/blob/master/.golangci.yml) agreed.

If we don't want to see the warning again, please configure lint tools in IDEA.

So I think there is no problem.

LinuxSuRen (approved):
/lgtm

PR #369, opened 2021-11-23
Return error when the config checking is failed in the controller cmd

LinuxSuRen:


cmd/controller/app/server.go
```diff
@@ -65,9 +65,11 @@ func NewControllerManagerCommand() *cobra.Command {
 		Use:   "controller-manager",
 		Short: `KubeSphere DevOps controller manager`,
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
-			if errs := s.Validate(); len(errs) == 0 {
-				err = Run(s, signals.SetupSignalHandler())
+			if errs := s.Validate(); len(errs) != 0 {
+				return utilerrors.NewAggregate(errs)
 			}
+
+			err = Run(s, signals.SetupSignalHandler())
```
JohnNiang, line 72: IMO, the code below could simplify `return err = Run(s, signals.SetupSignalHandler())`.

```go
err = Run(s, signals.SetupSignalHandler())
return
```

LinuxSuRen: I didn't see the big differences.

JohnNiang: Clear and consistent are big differences.

PR #375, opened 2021-11-25
Set the threshold of patch to 0% for codecov

LinuxSuRen:
For many cases, it's hard to add new test coverage. Such as the modification of markdown files. So, we don't need to let the ci to failed.

## References
* https://docs.codecov.com/docs/commit-status#patch-status


JohnNiang (commented):
/lgtm

PR #401, opened 2021-12-06
Enable build workflow on pushing to release branch

JohnNiang:
Enable build workflow on pushing to release branch, so that we can use docker image: `kubespheredev/devops-apiserver:release-3.2`.

/kind chore

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - release-*
       - tekton-support
       - jenkins-client-refactor
       - feat/*
@@ -37,6 +38,7 @@ on:
       - "PROJECT"
     branches:
       - 'master'
+      - 'release-*'
       - 'tekton-support'
       - 'jenkins-client-refactor'
       - 'feat/*'


LinuxSuRen:
Why do you suggest using `kubespheredev/devops-apiserver:release-3.2` instead of `kubesphere/devops-apiserver:v3.2.x`?

If there is anyone who wants to use a pre-release version. We can just create a pre-release. IMO, `release-3.2` this kind of version is always changing.

JohnNiang:
> Why do you suggest using `kubespheredev/devops-apiserver:release-3.2` instead of `kubesphere/devops-apiserver:v3.2.x`?
> 
> If there is anyone who wants to use a pre-release version. We can just create a pre-release. IMO, `release-3.2` this kind of version is always changing.

If some people want to verify fixes within branch release-3.2 as soon as possible, is there anyway to do that?

LinuxSuRen (commented):
Currently, we have the following image tag patter:
* `vx.x.x` represents the official image
* `master` built from the master branch, the purpose is letting developer test

If we allow publishing the container images by branch `release-*`. It will produce an image likes `xxx/xxx:release-3.2`. This will let users feel confused, if this is a release image if this image can be used in a production environment.

PR #426, opened 2021-12-27
Add an override flag to the tool cmd jwt

LinuxSuRen:
### What type of PR is this?

/kind feature

Users cannot set a customized token of the Jenkins if there's no the override flag.

Fixes #

You can use the following image to test it.
```
surenpi/devops-tools:dev-v3.2.1-f4b07eb
```

you can add the flag like this: `--override-jenkins-token=false`

```release-note
None
```


JohnNiang (approved):
/lgtm

PR #439, opened 2022-01-11
Support to verify SCM token directly instead of call Jenkins API

LinuxSuRen:
### What type of PR is this?

/kind feature

There are two big reasons why we should have this feature:

* Verify a SCM token has no relationship with Jenkins. So, it is not reasonable to call Jenkins API
* We should avoid use a plain token text as possible as we can

Currently, we use [go-scm](https://github.com/drone/go-scm) which support multiple git providers.

Fixes #

If you want to help to test it, please use the following temporary image:
```
kubespheredev/devops-apiserver:dev-v3.2.1-bdeb73d
```

image

Below is the test command:

```shell
curl -X POST 'http://ip:30880/kapis/devops.kubesphere.io/v1alpha3/scms/github/verify?secret=github&secretNamespace=default' \
Proxy-C> -H 'Proxy-Connection: keep-alive' \
> -H 'User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/96.0.4664.110 Safari/537.36' \
> -H 'content-type: application/json' \
> -H 'Accept: */*' \
> -H 'Referer: http://103.61.39.64:30880/test/clusters/default/devops/testbhhqx/pipelines' \
> -H 'Accept-Language: en-US,en;q=0.9,zh-CN;q=0.8,zh;q=0.7' \
> -H 'Cookie: lang=en; oAuthLoginInfo=; token=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NDE5MTU1NTgsImlhdCI6MTY0MTkwODM1OCwiaXNzIjoia3ViZXNwaGVyZSIsInN1YiI6ImFkbWluIiwidG9rZW5fdHlwZSI6ImFjY2Vzc190b2tlbiIsInVzZXJuYW1lIjoiYWRtaW4ifQ.f7pn9gxptym14NrfJkn4hZ_GrtLlAZ3_Hb_ZVP2oMew; expire=1641915558246; refreshToken=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJleHAiOjE2NDE5MjI3NTgsImlhdCI6MTY0MTkwODM1OCwiaXNzIjoia3ViZXNwaGVyZSIsInN1YiI6ImFkbWluIiwidG9rZW5fdHlwZSI6InJlZnJlc2hfdG9rZW4iLCJ1c2VybmFtZSI6ImFkbWluIn0.rbIPZDCEGsaq9rerz66oitJ8_sYjkokbJIoCMVxzxj0' \
> --compressed \
> --insecure ;
{
 "message": "ok",
 "code": 0,
 "errors": null
}
```

```release-note
Support to verify SCM token directly instead of call Jenkins API
```

### DIFF
--- a/pkg/client/git/message.go
+++ b/pkg/client/git/message.go
@@ -6,8 +6,6 @@ type VerifyResponse struct {
 	Message string `json:"message"`
 	// Code represents a group of cases
 	Code int `json:"code"`
-	// Errors contain all related errors
-	Errors []interface{} `json:"errors"`
 }
 
 func VerifyPass() *VerifyResponse {
@@ -16,17 +14,16 @@ func VerifyPass() *VerifyResponse {
 	}
 }
 
-func VerifyFailed(message string, code int, err error) *VerifyResponse {
+func VerifyFailed(message string, code int) *VerifyResponse {
 	return &VerifyResponse{
 		Message: message,
 		Code:    code,
-		Errors:  []interface{}{err},
 	}
 }
 
-func VerifyResult(message string, code int, err error) *VerifyResponse {
+func VerifyResult(err error, code int) *VerifyResponse {
 	if err == nil {
 		return VerifyPass()
 	}
-	return VerifyFailed(message, code, err)
+	return VerifyFailed(err.Error(), code)
 }
--- a/pkg/client/git/message.go
+++ b/pkg/client/git/message.go
@@ -1,0 +1,32 @@
+package git
+
+// VerifyResponse represents a response of SCM auth verify
+type VerifyResponse struct {
+	// Message is the detail of the result
+	Message string `json:"message"`
+	// Code represents a group of cases
+	Code int `json:"code"`
+	// Errors contain all related errors
+	Errors []interface{} `json:"errors"`
+}
+
+func VerifyPass() *VerifyResponse {
+	return &VerifyResponse{
+		Message: "ok",
+	}
+}
+
+func VerifyFailed(message string, code int, err error) *VerifyResponse {
+	return &VerifyResponse{
+		Message: message,
+		Code:    code,
+		Errors:  []interface{}{err},
+	}
+}
+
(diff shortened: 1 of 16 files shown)


pkg/kapis/devops/v1alpha3/register.go
```diff
@@ -64,6 +66,10 @@ func AddToContainer(container *restful.Container, devopsClient devopsClient.Inte
 func registerRoutes(devopsClient devopsClient.Interface, k8sClient k8s.Client, client client.Client, ws *restful.WebService) {
 	handler := newDevOpsHandler(devopsClient, k8sClient)
 	registerRoutersForCredentials(handler, ws)
+	registerRoutersForCredentials(handler, ws)
```
JohnNiang, line 69: Duplicate registering routers for credentials, please remove it.

pkg/client/git/message.go
```diff
@@ -0,0 +1,32 @@
+package git
+
+// VerifyResponse represents a response of SCM auth verify
+type VerifyResponse struct {
+	// Message is the detail of the result
+	Message string `json:"message"`
+	// Code represents a group of cases
+	Code int `json:"code"`
+	// Errors contain all related errors
+	Errors []interface{} `json:"errors"`
```
JohnNiang, line 10: How about changing type of `Errors` to `[]error` instead of `[]interface{}`?

JohnNiang (changes requested):
All changes look good to me, but two places below. Please take a look at it.

pkg/kapis/devops/v1alpha3/register.go
LinuxSuRen: Sure.

JohnNiang (approved):
/lgtm

PR #459, opened 2022-02-08
Re-orgnize the controllers package

LinuxSuRen:
### What type of PR is this?

/kind chore

It can help contributors to understand all the controllers.

Fixes #

I deleted an useless CRD: `fake`

```release-note
None
```


pkg/api/devops/v1alpha3/register.go
```diff
@@ -0,0 +1,23 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha3
+
+func init() {
```
JohnNiang, line 19: I don't think we need to register schemes here. Please see 

https://github.com/kubesphere/ks-devops/blob/c6d6f2a3b4abad901727ca468175cb4a3441de58/pkg/api/devops/v1alpha3/pipeline_types.go#L259-L261

https://github.com/kubesphere/ks-devops/blob/c6d6f2a3b4abad901727ca468175cb4a3441de58/pkg/api/devops/v1alpha3/pipelinerun_types.go#L303-L305

https://github.com/kubesphere/ks-devops/blob/c6d6f2a3b4abad901727ca468175cb4a3441de58/pkg/api/devops/v1alpha3/devopsproject_types.go#L71-L73

JohnNiang (changes requested):

pkg/api/devops/v1alpha3/register.go
LinuxSuRen: I have removed all the `Registers` from each go source file. It is not a good practice to have many `init` functions.

JohnNiang: Those init functions are generated by `kubebuilder` when creating that API. You can see the official sample layout: https://github.com/kubernetes-sigs/kubebuilder/tree/master/docs/book/src/cronjob-tutorial/testdata/project/api/v1.

BTW, is there any reference to show many `init` functions are not a good practice? You can check this: https://go.dev/doc/effective_go#init as well.

LinuxSuRen: It's more clear if we put all register together. And it's not necessary to have too many `init` functions that have the same purpose. Of course, this is not a big deal if you insist on negative feedback.

JohnNiang (approved):
/lgtm

Issue #63, opened 2021-07-19
Argo CD integration request


LinuxSuRen:
Currently, the KubeSphere community provides a CD solution that takes advantage of [kubernetes-cd-plugin](https://github.com/jenkinsci/kubernetes-cd-plugin). But this plugin is out-of-date and lacks maintenance. So I highly don't recommend you continue to use it.

We're looking forward to a better CD solution.

I don't have enough information for this. But [Argo CD](https://github.com/argoproj/argo-cd/) is a very popular GitOps project. Please contact us, if there's anyone who wants to help with this.

/help
/stage alpha

LinuxSuRen:
We can expand the `devopsproject` with [the project](https://github.com/argoproj/argo-cd/blob/d0cbb1597ca8b93cae081e01c07c5c2e74f580d4/pkg/apis/application/v1alpha1/app_project_types.go#L38) of Argo CD.
```yaml
apiVersion: devops.kubesphere.io/v1alpha4
kind: devopsproject
spec:
  kind: argo-project
  resource:
    apiVersion: argoproj.io/v1alpha1
    kind: appproject
    spec:
      xxx: xx
```

Application sample:
```yaml
apiVersion: devops.kubesphere.io/v1alpha1
kind: Application
metadata:
  name: xxx
spec:
    kind: argocd-app|fluxcd|argocd-appset|argocd-app-patch
    source:
      repoURL: https://xx
      path: "/"
      targetRevision: ""
    destination:
      server: xxx
      namespace: xxx
      name: xx
    syncPolicy:
      automated:
        prune: true
        allowEmpty: true
    resource:
      apiVersion: argoproj.io/v1alpha1
      kind: ApplicationSet
      metadata:
        name: ks-devops-appset
        namespace: argocd
      spec:
        generators:
```

LinuxSuRen:
The UI design might be have a draft in this week.

LinuxSuRen:
Let's assume there are two environments: `int` and `prod`, and one project [ks-releaser](https://github.com/kubesphere-sigs/ks-releaser).

Use cases:

* Once a PR got merged, deploy the latest development version to the `int` environment
  * In this case, we can add a stage at the end of the pipeline. Change the image tag or digest from the [ks-infra](https://github.com/kubesphere-sigs/ks-infra/). We can commit and push it directly, or just file a PR for it.
* Once a release was created, deploy the latest release version to the `prod` environment
[...]

PR #463, opened 2022-02-11
Add Argo CD support to DevOpsProject CRD

LinuxSuRen:
### What type of PR is this?

/kind feature

This is part of #63 

In this PR, I put the `AppProject` into the `DevOpsProject`. So, `DevOpsProject` will be responsible to maintain `AppProject`.

Fixes #

This is the temporary image that you can test against this PR.
```
kubespheredev/devops-controller:dev-v3.2.1-35bf471
```

```release-note
None
```


controllers/argocd/project_controller.go
```diff
@@ -0,0 +1,187 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// Reconciler is the reconciler of the DevOpsProject with Argo AppProject
+type Reconciler struct {
+	client.Client
+	log      logr.Logger
+	recorder record.EventRecorder
+}
+
```
JohnNiang, line 43: Is the configuration of RBAC ignored here?

Please refer to https://github.com/kubesphere/ks-devops/blob/c78b2f252e37030b6241ff53a2e48d258ee0a1e2/controllers/template/template_controller.go#L36-L37.

pkg/api/devops/v1alpha3/devopsproject_types.go
```diff
@@ -23,20 +23,121 @@ import (
...
 	DevOpsProjectPrefix            = "devopsproject.devops.kubesphere.io/"
+	DevOpsProjectFinalizerName     = "devopsproject.finalizers.kubesphere.io"
 	DevOpeProjectSyncStatusAnnoKey = DevOpsProjectPrefix + "syncstatus"
 	DevOpeProjectSyncTimeAnnoKey   = DevOpsProjectPrefix + "synctime"
-	DevOpeProjectSyncMsgAnnoKey    = DevOpsProjectPrefix + "syncmsg"
 )
 
 // DevOpsProjectSpec defines the desired state of DevOpsProject
 type DevOpsProjectSpec struct {
+	Argo *Argo `json:"argo,omitempty"`
```
JohnNiang, line 38: IMO, we should create a new version(like v1alpha4) of DevOpsProject instead of editing previous version(v1alpha3).

JohnNiang (changes requested):

pkg/api/devops/v1alpha3/devopsproject_types.go
LinuxSuRen: It needs too much work if we have a new version.

JohnNiang: Of cause, I understand that it's not easy. But we should consider a scenario: upgrading from a previous version. If we do like this, the feature will not work as we want. Maybe we need to do more work to fix compatibility issues.

LinuxSuRen: > we should consider a scenario: upgrading from a previous version

There is no compatible issue for this PR. Because it only adds some new fields instead of modifying anything.

controllers/argocd/project_controller.go
LinuxSuRen: Good catch!

JohnNiang (approved):
I'm not sure what the pitfalls of adding new field to an existing type can cause. So I don't think this is a block reason here.

/lgtm

PR #469, opened 2022-02-23
Support sync kubesphere multi-cluster to argo multi-cluster

LinuxSuRen:
/kind feature

this is part of #63 

With help from this PR, users don't need to add the [ArgoCD cluster](https://argo-cd.readthedocs.io/en/stable/user-guide/commands/argocd_cluster/) manually. They just need to follow [this document](https://kubesphere.com.cn/docs/multicluster-management/introduction/overview/) to add a member cluster, the controller will add the corresponding ArgoCD cluster.

Fixes #

You can use the following image if you want to help to test this PR.
```
surenpi/devops-controller:dev-v3.2.1-50bb80b
```

```release-note
None
```

### DIFF
--- a/controllers/argocd/multi-cluster-controller.go
+++ b/controllers/argocd/multi-cluster-controller.go
@@ -1,0 +1,223 @@
+/*
+Copyright 2022 The KubeSphere Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


controllers/argocd/multi-cluster-controller.go
```diff
@@ -0,0 +1,223 @@
...
+	k8sutil.SetOwnerReference(existingArgoCluster, ownerRef)
+	err = r.Update(ctx, existingArgoCluster)
+	return
+}
+
+// let's assume there is only one ArgoCD existing
+// TODO consider to have a better solution to figure the namespace of ArgoCD
+func (r *MultiClusterReconciler) findArgoCDNamespace() string {
+	secretList := &v1.SecretList{}
+	if err := r.List(context.Background(), secretList); err == nil {
```
JohnNiang, line 110: It's better to use `r.Get(...)` to find `argocd-secret` Secret instead of listing and foreach. Please correct me if I missed something.

LinuxSuRen: The problem is that I don't know the secret namespace. So it looks like no way to use `Get`.

JohnNiang (approved):
/lgtm

Issue #475, opened 2022-02-24
Add permission of the new API '/scms/'


LinuxSuRen:
### What is version of KubeSphere DevOps has the issue?

v3.3

### Describe this feature

Add some new APIs to remove the rely on Jenkins.

LinuxSuRen:
As we can see, the corresponding permission exists in a global role. And I have tested it manually, it works well.

I created a custom role that only supports creating Pipelines. A user with that role can list the organizations and repositories.

See also https://github.com/kubesphere/ks-installer/blob/1cf512276d219d98eea97c6096bcc1edd5ba1da1/roles/ks-core/prepare/files/ks-init/role-templates.yaml#L216

/close

PR #476, opened 2022-02-24
Add API of listing organizations and repositories

LinuxSuRen:
### What type of PR is this?

/kind feature

Listing the organizations and repositories should not rely on the Jenkins.

Fixes #

  see #475

```release-note
None
```


pkg/apiserver/apiserver.go
```diff
@@ -183,6 +183,7 @@ func (s *APIServer) Run(stopCh <-chan struct{}) (err error) {
 	}()
 
 	klog.V(0).Infof("Start listening on %s", s.Server.Addr)
+	klog.V(0).Infof("Open the swagger-ui from http://localhost%s/apidocs", s.Server.Addr)
```
JohnNiang, line 186: It would be better if we print swagger link like <http://localhost:9090/apidocs/?url=http://localhost:9090/apidocs.json/>. It will load swagger API configuration automatically.

pkg/kapis/devops/v1alpha3/scm/scmhandler.go
```diff
@@ -22,43 +22,134 @@ import (
...
 			code = resp.Status
+
+			for i := range allRepos {
+				repo := allRepos[i]
+				if repo.Namespace == org {
+					repos = append(repos, repo)
+				}
+			}
+		} else {
+			code = 101
```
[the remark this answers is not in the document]
LinuxSuRen: You caught me. Those codes do not have a specific meaning. Any suggestions here?

pkg/apiserver/apiserver.go
LinuxSuRen: done

JohnNiang (approved):
/lgtm

PR #478, opened 2022-02-25
Move CRD application into a new group gitops

LinuxSuRen:
/kind chore

Create a new CRD group `gitops` to store all things related to GitOps.

Why do I create a new CRD group? Because always putting new CRDs into the v1alpha1 might letting users or contributors feel confused. So, in the future. It is best to expand this project with a new group instead of putting it into the core part.

Fixes #

```release-note
None
```


JohnNiang (approved):
/lgtm

PR #482, opened 2022-03-03
Add a flag to support skip Jenkins connection in the controller cmd

LinuxSuRen:
### What type of PR is this?

/kind feature

In some cases, a developer might only want to test a particular feature instead of the full features. So, it would be better if we can ignore the unnecessary feature.

Fixes #

```release-note
None
```


JohnNiang (approved):
/lgtm

PR #487, opened 2022-03-07
Add missing git secret support which belong to ks type

LinuxSuRen:
/kind bug

See the following screenshot:

image
image

Currently, this project does not use the original secret type.

Fixes #

this is the upstream of https://github.com/kubesphere/console/pull/2943

```release-note
None
```

### DIFF
--- a/pkg/client/git/client.go
+++ b/pkg/client/git/client.go
@@ -82,10 +82,12 @@ func (c *ClientFactory) getTokenFromSecret(secretRef *v1.SecretReference) (token
 	}
 
 	switch gitSecret.Type {
-	case v1.SecretTypeBasicAuth:
+	case v1.SecretTypeBasicAuth, "credential.devops.kubesphere.io/basic-auth":
 		token = string(gitSecret.Data[v1.BasicAuthPasswordKey])
 	case v1.SecretTypeOpaque:
 		token = string(gitSecret.Data[v1.ServiceAccountTokenKey])
+	case "credential.devops.kubesphere.io/secret-text":
+		token = string(gitSecret.Data["secret"])
 	}
 	return
 }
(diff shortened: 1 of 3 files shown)


pkg/client/git/client.go
```diff
@@ -82,10 +82,12 @@ func (c *ClientFactory) getTokenFromSecret(secretRef *v1.SecretReference) (token
 	}
 
 	switch gitSecret.Type {
-	case v1.SecretTypeBasicAuth:
+	case v1.SecretTypeBasicAuth, "credential.devops.kubesphere.io/basic-auth":
```
JohnNiang, line 85: It would be better if we use 
https://github.com/kubesphere/ks-devops/blob/69dda0b28cad59fe187c5d9bce972fee3ebfb7f8/pkg/api/devops/v1alpha3/credential_types.go#L34

pkg/client/git/client.go
```diff
@@ -82,10 +82,12 @@ func (c *ClientFactory) getTokenFromSecret(secretRef *v1.SecretReference) (token
...
 		token = string(gitSecret.Data[v1.BasicAuthPasswordKey])
 	case v1.SecretTypeOpaque:
 		token = string(gitSecret.Data[v1.ServiceAccountTokenKey])
+	case "credential.devops.kubesphere.io/secret-text":
```
JohnNiang, line 89: I suggest using 
https://github.com/kubesphere/ks-devops/blob/69dda0b28cad59fe187c5d9bce972fee3ebfb7f8/pkg/api/devops/v1alpha3/credential_types.go#L58
instead.

pkg/client/git/client.go
LinuxSuRen: IMO, it's not ok to let a git client use the package `ks-devops/pkg/api/devops/v1alpha3`

JohnNiang (approved):
/lgtm

PR #489, opened 2022-03-08
Add required field into template parameter

JohnNiang:
### What type of PR is this?

/kind feature

As mentioned in the title.

Fixes #

```release-note
None
```


pkg/api/devops/v1alpha3/template_types.go
```diff
@@ -48,6 +48,10 @@ type TemplateParameter struct {
 	//+optional
 	Description string `json:"description,omitempty"`
 
+	// Required indicates if this parameter is mandatory.
+	//+optional
+	Required bool `json:"required"`
```
LinuxSuRen, line 53: I'm wondering if this field needs `omitempty` or not. What do you think about it?

JohnNiang: I agree with you. Good catch!

LinuxSuRen (commented):
/lgtm
/approve

PR #491, opened 2022-03-08
Remove s2i reconciler from ks-devops

### LINKED ISSUE #450: Request to remove S2I reconcilers from ks-devops
As we all know, KubeSphere provides Source-to-Image (S2I) and Binary-to-Image (B2I) features to automate image building and pushing and application deployment. See [here](https://kubesphere.io/docs/project-user-guide/image-builder/s2i-and-b2i-webhooks/) for more.

But we already have an [S2I operator](https://github.com/kubesphere/s2ioperator) to take charge of this feature. Of course, I have tested it by disabling DevOps Controller, all functions of S2I and B2I work as normal.

Due to coupling with S2I, we can not remove S2I components now completely, such as S2I types and handlers. However, we could remove S2I reconcilers first or disable them by default.

/kind cleanup

mangoGoForward:
### What type of PR is this?

/kind cleanup

Remove s2i reconciler from ks-devops.

Fixes #450 

/cc @kubesphere/sig-devops 

```release-note
None
```


JohnNiang (approved):
/lgtm

cmd/controller/app/controllers.go
```diff
@@ -165,19 +162,6 @@ func getAllControllers(mgr manager.Manager, client k8s.Client, informerFactory i
...
 			return err
 		},
-		"s2ibinary": func(mgr manager.Manager) error {
-			return mgr.Add(s2ibinary.NewController(client.Kubernetes(),
-				client.KubeSphere(),
-				kubesphereInformer.Devops().V1alpha1().S2iBinaries(),
-				s3Client,
-			))
-		},
-		"s2irun": func(mgr manager.Manager) error {
```
LinuxSuRen, line 175: You already removed this part. So, what about the code base of `s2irun.NewS2iRunController`. Do we still need to keep them? If you think it should be kept, it would be better if you could let us know the reason.

LinuxSuRen (changes requested):

cmd/controller/app/controllers.go
LinuxSuRen: By the way, should we remove the CRDs as well? And please don't forget to test it.

mangoGoForward: Those controllers could be removed, but the CRDs will recreate when we execute the command of `make manifests generate`, the `s2ibinary type` can not remove yet.

JohnNiang: Yup, currently, the s2i types cannot be removed. But s2i controllers could be removed safely because we would never run them forever in devops-controller.

JohnNiang (approved):
/lgtm

LinuxSuRen:
Below are the images of this PR:
···
kubespheredev/devops-controller:dev-v3.2.1-ada16ac
···

···
kubespheredev/devops-apiserver:dev-v3.2.1-ada16ac
···

LinuxSuRen (commented):
It looks ok when I try to replace the images then test it.

/lgtm
/approve

PR #495, opened 2022-03-10
Provider the APIs for the GitRepository custom resource

LinuxSuRen:
### What type of PR is this?

/kind feature

As I mentioned in the title, this PR will add the APIs of the custom resource GitRepository.

And I change the version of `GitRepository` and `Addon` to `v1alpha3`. Because it belongs to the core part. Currently, all the core part are using `v1alpha3`.

Fixes #

```release-note
None
```


pkg/kapis/devops/v1alpha3/scm/scmhandler.go
```diff
@@ -29,11 +29,15 @@ import (
 // handler holds all the API handlers of SCM
 type handler struct {
 	k8sClient client.Client
+	client.Client
```
JohnNiang, line 32: IMO, `k8sClient client.Client` and `client.Client` are the same type. How about removing one of them?

JohnNiang (changes requested):

pkg/kapis/devops/v1alpha3/scm/gitrepository_handler.go
```diff
@@ -0,0 +1,96 @@
...
+
+	err := req.ReadEntity(patchRepo)
+	if err == nil {
+		err = h.Get(ctx, types.NamespacedName{
+			Namespace: namespace,
+			Name:      repoName,
+		}, repo)
+		if err == nil {
+			repo.Spec = patchRepo.Spec
+			err = h.Update(ctx, repo)
```
JohnNiang, line 76: Do we really update spec only? What about others? If it's true, it is recommended to mention it in the API doc.

BTW, This doubt won't prevent others from merging this PR.

LinuxSuRen: I'm not sure about that. Consider there is no UI that allows us to test it, I'm not able to think up enough details on that. I prefer to update the API once more details become clear.

JohnNiang (commented):
/lgtm

PR #505, opened 2022-03-21
Add missing RBAC permission claim

LinuxSuRen:
### What type of PR is this?

/kind chore

As I mentioned in the title. And we should not give too open permission to ks-devops. So, the next step is to change the `cluster-role` of [this file](https://github.com/kubesphere-sigs/ks-devops-helm-chart/blob/master/charts/ks-devops/templates/serviceaccount.yaml).

Fixes #

```release-note
None
```

### DIFF
--- a/controllers/argocd/application_controller.go
+++ b/controllers/argocd/application_controller.go
@@ -100,6 +100,11 @@ func createUnstructuredApplication(app *v1alpha1.Application) (result *unstructu
 		argoApp.Spec.Project = "default"
 	}
 
+	// application destination can't have both name and server defined
+	if argoApp.Spec.Destination.Name != "" && argoApp.Spec.Destination.Server != "" {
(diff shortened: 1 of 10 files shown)


controllers/argocd/application_controller.go
```diff
@@ -100,6 +100,11 @@ func createUnstructuredApplication(app *v1alpha1.Application) (result *unstructu
 		argoApp.Spec.Project = "default"
 	}
 
+	// application destination can't have both name and server defined
+	if argoApp.Spec.Destination.Name != "" && argoApp.Spec.Destination.Server != "" {
```
LinuxSuRen, line 104: I made this small change here. It has nothing about this PR. I can file a new PR against this if you think it should not be here.

JohnNiang (approved):
/lgtm

PR #515, opened 2022-03-24
Sort by creationTimestamp in descending order by default

JohnNiang:
### What type of PR is this?

/kind bug

Sort by creationTimestamp in descending order by default while using default comparison function. Or it will produce messy sort of result when searching a lot of resources.

Fixes #

```release-note
None
```

/cc @kubesphere/sig-devops


JohnNiang:
/close

I found that the default sort behavior was correct. So I'm going to close this PR.

PR #540, opened 2022-04-08
Bump golangci lint action to v3

JohnNiang:
### What type of PR is this?

/kind bug

This may resolve the lint errors like below:

image

The result is as below:

image

Fixes https://github.com/kubesphere/ks-devops/actions/runs/2114131090

/cc @kubesphere/sig-devops 

```release-note
None
```

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -49,7 +49,7 @@ jobs:
             exit 1
           fi
       - name: golint
-        uses: golangci/golangci-lint-action@v2
+        uses: golangci/golangci-lint-action@v3
         with:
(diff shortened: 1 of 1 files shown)


LinuxSuRen (commented):
hi @JohnNiang , I found a potential useful feature on v3.1.0 of the golangci-lint-action.

* Allow to disable caching completely by @tdabasinskas in https://github.com/golangci/golangci-lint-action/pull/351

see also https://github.com/golangci/golangci-lint-action/releases/tag/v3.1.0

So, I'm thinking if we can use v3.1.0 instead.

LinuxSuRen (approved):

PR #556, opened 2022-04-20
fix: ParameterReference apiGroup empty

lxm:
/kind bug

pipeline save will get a error when apiGroup is empty.

https://github.com/kubernetes/kubernetes/issues/79354

Although the comment says it's optional, we'll still get an error if we pass null because there's no omitempty 

```release-note
None
```

### DIFF
--- a/config/crd/bases/devops.kubesphere.io_pipelineruns.yaml
+++ b/config/crd/bases/devops.kubesphere.io_pipelineruns.yaml
@@ -375,7 +375,6 @@ spec:
                             type:
                               type: string
                           required:
-                          - isQuoted
                           - name
                           - type
                           type: object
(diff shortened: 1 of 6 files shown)


LinuxSuRen (changes requested):
hi @lxm , just a friendly tip. The CI process is failed.

lxm:
> CI

right, but I think the CI failure has nothing to do with this PR. We need another PR to fix the security issue.

LinuxSuRen:
> I think the CI failure has nothing to do with this PR. We need another PR to fix the security issue.

Fair enough. Let's just hold this PR until the blocked issues are fixed.
/hold

LinuxSuRen:
Now @lxm you can rebase the current branch to upstream/master. I belive the CI will pass.

PR #655, opened 2022-06-09
Condition is always 'true' because 'user' is always not 'nil'

nuclearwu:
### What type of PR is this?
/kind cleanup

Condition is always 'true' because 'user' is always not 'nil'

Fixes #

None
```release-note
None
```

### DIFF
--- a/pkg/kapis/devops/v1alpha3/pipelinerun/handler.go
+++ b/pkg/kapis/devops/v1alpha3/pipelinerun/handler.go
@@ -138,7 +138,7 @@ func (h *apiHandler) createPipelineRun(request *restful.Request, response *restf
 	}
 	// create PipelineRun
 	pr := CreatePipelineRun(&pipeline, &payload, scm)
-	if user != nil && user.GetName() != "" {
+	if user.GetName() != "" {
 		pr.GetAnnotations()[v1alpha3.PipelineRunCreatorAnnoKey] = user.GetName()
 	}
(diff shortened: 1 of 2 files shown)


pkg/kapis/devops/v1alpha3/pipelinerun/handler.go
```diff
@@ -138,7 +138,7 @@ func (h *apiHandler) createPipelineRun(request *restful.Request, response *restf
 	}
 	// create PipelineRun
 	pr := CreatePipelineRun(&pipeline, &payload, scm)
-	if user != nil && user.GetName() != "" {
```
LinuxSuRen, line 141: Thanks for improving this line. I'm thinking if it's possible to add a unit test for this case. I know that line 133 already makes sure the variable `user` is not nil. But in case other contributors forget that a unit test might be the perfect solution.

This is only a recommendation. We can merge this PR without new unit tests. But please see also the following examples if you want to add new unit tests.

https://github.com/kubesphere/ks-devops/blob/master/pkg/kapis/devops/v1alpha3/steptemplate/route_test.go

LinuxSuRen:
> /assign @LinuxSuRen

hi @nuclearwu , thanks for your effort. Glad to see your second PR. But I'd like to suggest you use the following comment to ping the DevOps team instead of the individual.

`/cc @kubesphere/sig-devops `

LinuxSuRen (approved):
This is really cool! Thanks for your great effort!

Issue #718, opened 2022-07-07
[FluxCD Integration] Add FluxCD Application Management


chengleqi:
Create a new application-controller to manage FluxCD Application which include both Kustomization and HelmRelease.

/kind feature

chengleqi:
[WIP] https://github.com/chengleqi/ks-devops/tree/add-app-ctrl

chengleqi:
fixed by #743

PR #743, opened 2022-07-18
Add FluxCD Application Management

### LINKED ISSUE #716: [FluxCD Integration] Support to manage HelmRelease Application Type
Target:  sync the underlying helmreleases.helm.toolkit.fluxcd.io CRD with a New gitops.kubesphere.io/valpha2/Application CRD

Task:

- [ ] #679

### LINKED ISSUE #717: [FluxCD Integration] Update the gitops.kubesphere.io/v1alpha1 CRD to Support FluxCD Application
/kind feature

chengleqi:
### What type of PR is this?
/kind feature
/kind design

- Extend the gitops.kubesphere.io/v1alpha1/Application CRD to Support FluxApp.
- Add fluxcd/**application-controller** to maintain the fluxcd [Kustomization](https://fluxcd.io/docs/components/kustomize/kustomization/) and [HelmRelease](https://fluxcd.io/docs/components/helm/helmreleases/).
- Update the config/rbac/role.yaml
- Add sample files in config/samples/gitops directory.

Fixes #717 , #718 

- [x] Already added the RBAC markers for the new controllers

```release-note
None
```


chengleqi:
I encountered a strange problem when I wrote unit test due to the version of controller-runtime.

I use code like this `r.Client.List(..., &unstructured.UnstructuredList{}, ...)`, It run correctly when the Client is RestClient, but It'll run into error when the client is fake client in the unit test scenario.

I found the root cause is [this issue](https://github.com/kubernetes-sigs/controller-runtime/issues/702), but the [PR which fix this issue](https://github.com/kubernetes-sigs/controller-runtime/pull/1521) was merged only in the 0.9.x version of controller-runtime, I tried to update the controller-runtime v0.6.3 => v0.9.7 but It caused too many changes.

So I Added code manually like this in the unit test:

```golang
schema.AddKnownTypeWithName(apischema.GroupVersionKind{
        Group:   "helm.toolkit.fluxcd.io",
        Version: "v2beta1",
        Kind:    "HelmReleaseList",
}, &unstructured.UnstructuredList{})
```

And the fluxcd project actually provide the [module](github.com/fluxcd/helm-controller/api) which include the [HelmRelease type](https://github.com/fluxcd/helm-controller/blob/a2147639031eb1362360b52c6f596c066270ecc4/api/v2beta1/helmrelease_types.go#L923) for convenience. But It also need higher version of controller-runtime.

Is it possible to update the controller-runtime version when it arrive version 1.0 or LTS ?

chengleqi:
When I delete the gitops.kubesphere.io/Application I want to delete all HelmReleases belong to it. I can implement it just set the all HelmReleases' `OwnerReferences` to the Application. And I also can set the Finalizer on Application to deal with it. 

Is there any different between the `OwnerReferences` solution and the `Finalizer` solution in this situation?

LinuxSuRen:
> Is there any different between the `OwnerReferences` solution and the `Finalizer` solution in this situation?

I'm not sure if I understand it correctly. But it might be helpful for you.

IMO, the `OwnerReferences` is a built-in controller for resource collection. It could help you to delete Kubernetes CR (custom resource) instead of others (Jenkins jobs .e.g.). And for this case, the child resource does not have its own life cycle.

The `Finalizer` could let you delete any resources that you want.

See also https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/

chengleqi:
> > Is there any different between the `OwnerReferences` solution and the `Finalizer` solution in this situation?
> 
> I'm not sure if I understand it correctly. But it might be helpful for you.
> 
> IMO, the `OwnerReferences` is a built-in controller for resource collection. It could help you to delete Kubernetes CR (custom resource) instead of others (Jenkins jobs .e.g.). And for this case, the child resource does not have its own life cycle.
> 
> The `Finalizer` could let you delete any resources that you want.
> 
> See also https://kubernetes.io/blog/2021/05/14/using-finalizers-to-control-deletion/

Thanks a lot, this is very helpful.

config/samples/gitops/fluxcd-application-helmrelease.yaml
```diff
@@ -0,0 +1,48 @@
...
+            chart: ./helm-chart
+            version: "0.1.0"
+            valuesFiles:
+              - ./helm-chart/values.yaml
+              - ./helm-chart/aliyun-values.yaml
+            reconcileStrategy: Revision
+          template:
+          deploy:
+            - destination:
+                kubeConfig:
```
lxm, line 30: should give a sample kubeconfig ref here

config/samples/gitops/fluxcd-application-helmtemplate.yaml
```diff
@@ -0,0 +1,25 @@
...
+  fluxApp:
+    spec:
+      source:
+      config:
+        helmRelease:
+          chart:
+          template: my-devops-projecthmhx2-chengleqi-test-helm
+          deploy:
+            - destination:
+                kubeConfig:
```
lxm, line 17: same above

config/samples/gitops/fluxcd-application-kustomization.yaml
```diff
@@ -0,0 +1,22 @@
...
+    spec:
+      source:
+        sourceRef:
+          kind: GitRepository
+          name: fluxcd-gitee-repo
+          namespace: my-devops-projecthmhx2
+      config:
+        kustomization:
+          - destination:
+              kubeConfig:
```
lxm, line 18: same above

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,426 @@
...
+}
+
+// Reconcile sync the Application with underlying FluxCD HelmRelease and FluxCD Kustomization CRD
+func (r *ApplicationReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
+	ctx := context.Background()
+	r.log.Info(fmt.Sprintf("start to reconcile application: %s", req.String()))
+
+	app := &v1alpha1.Application{}
+	// only reconcile fluxcd application
+	app.Spec.Kind = v1alpha1.FluxCD
```
lxm, line 57: I don't quite understand this line.
Do you want to express the following meaning?
```
if  strings.compare(app.Spec.kind, v1alpha1.FluxCD) != 0 {
    return 
}
```

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,426 @@
...
+		}
+	}
+	_ = argocd.SetNestedField(kus.Object, deploy.Destination.KubeConfig, "spec", "kubeConfig")
+	_ = unstructured.SetNestedField(kus.Object, deploy.Destination.TargetNamespace, "spec", "targetNamespace")
+}
+
+func isHelmOrKustomize(app *v1alpha1.Application) AppType {
+	if app.Spec.FluxApp.Spec.Config.HelmRelease != nil {
+		return HelmRelease
+	}
```
lxm, line 282: Need check if `Kustomization` not equal to nil ? and then return an error ?

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,426 @@
...
+func (r *ApplicationReconciler) GetGroupName() string {
+	return controllerGroupName
+}
+
+// GetName returns the name of this reconciler
+func (r *ApplicationReconciler) GetName() string {
+	return "FluxApplicationReconciler"
+}
+
+func getHelmReleaseName(targetNamespace string) string {
```
lxm, line 296: What is the purpose of this method? It just retruns the passed value directly

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,426 @@
...
+	return targetNamespace
+}
+
+func getKustomizationName(targetNamespace string) string {
```
lxm, line 300: same above

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,426 @@
...
+	return targetNamespace
+}
+
+func getHelmTemplateName(ns, name string) string {
+	return ns + "-" + name
+}
+
+func getKusManagerName(ns, name string) string {
```
lxm, line 308: KsManagerName?

pkg/api/gitops/v1alpha1/application.go
```diff
@@ -17,12 +17,694 @@ limitations under the License.
 package v1alpha1
 
 import (
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
```
lxm, line 23: Is it possible to import this structure from flux or helm-controller repo?

lxm (changes requested):

config/samples/gitops/fluxcd-application-kustomization.yaml
chengleqi: It can only reconcile local cluster flux app for now if we don't do some thing like below manually.
`kubectl create secret generic prod-kubeconfig --from-file=value.yaml=./kubeconfig`

And I'll write `multi-cluster-controller` in the next few days to reconcile the kubeconfig. Then I'll add the sample kubeconfig.

controllers/fluxcd/application-controller.go
chengleqi: Yes. I'll fix it.

controllers/fluxcd/application-controller.go
chengleqi: I'll change these function in the future after I write the `multi-cluster-controller`.

Here is the named design. The name of `HelmRelease` and `Kustomization` will be `clustername-targetnamespace`.

Application命名设计

controllers/fluxcd/application-controller.go
chengleqi: I' ll change this.

controllers/fluxcd/application-controller.go
LinuxSuRen: Do we have the corresponding unit tests for this?

controllers/fluxcd/application-controller.go
LinuxSuRen: It would be clear if you put a `TODO` comment here.

chengleqi: Thanks for your advice.

controllers/fluxcd/application-controller_test.go
```diff
@@ -0,0 +1,683 @@
...
+			args: args{
+				req: ctrl.Request{
+					NamespacedName: types.NamespacedName{
+						Namespace: "fake-ns",
+						Name:      "fake-app",
+					},
+				},
+			},
+			wantErr: assert.NoError,
+		},
```
chengleqi, line 112: I actually wrote unit test for this but these test case has not find this bug. @LinuxSuRen

LinuxSuRen: Yes, it would be great if we could cover that case in unit tests.

controllers/fluxcd/application-controller.go
lxm: Using a delimiter in a name may cause other problems, such as the length limit of the name. Can these contents be stored with annotations?

chengleqi: Yes. I run into length limit problem once. Maybe I can use `SHA` to avoid the naming conflict and leave these information into annotations.

pkg/api/gitops/v1alpha1/application.go
lxm: Changing to latest version of controller-runtime seems to be a very important thing. 
Is there an updated detailed description of the work required for upgrading @kubesphere/sig-devops

LinuxSuRen: I prefer to do the upgrade work in a separate PR.

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,425 @@
...
+	return "FluxApplicationReconciler"
+}
+
+func getHelmReleaseName(deploy *v1alpha1.Deploy) string {
+	// host cluster
+	if deploy.Destination.KubeConfig == nil {
+		return deploy.Destination.TargetNamespace
+	}
+	// member cluster
+	return deploy.Destination.KubeConfig.SecretRef.Name + "-" + deploy.Destination.TargetNamespace
```
lxm, line 309: This is still named using the separator

lxm (changes requested):

controllers/fluxcd/application-controller.go
chengleqi: I'll fix this.

lxm:
> @chengleqi: Cannot trigger testing until a trusted user reviews the PR and leaves an `/ok-to-test` message.

@kubesphere/sig-devops @LinuxSuRen

controllers/fluxcd/application-controller_test.go
```diff
@@ -310,17 +310,24 @@ func TestApplicationReconciler_reconcileApp(t *testing.T) {
 	helmDeploy := helmApp.Spec.FluxApp.Spec.Config.HelmRelease.Deploy[0]
 	setFluxHelmReleaseFields(fluxHR, fluxHelmChart, helmDeploy)
 	fluxHR.SetNamespace(helmApp.GetNamespace())
-	fluxHR.SetName(getHelmReleaseName(helmDeploy))
+	fluxHR.SetName(helmApp.GetName() + "abcde")
```
lxm, line 313: why `abcde` suffix?

lxm (changes requested):

lxm:
@kubesphere/sig-devops

LinuxSuRen:
> /test ?

You can retrigger the actions once you close and reopen this PR.

chengleqi:
@kubesphere/sig-devops

lxm:
while controller-runtime version has been updated to v0.12.3 with #788.
some previously change requests in the PR can be performed, such as data struct import method

pkg/api/gitops/v1alpha1/application.go
lxm: we can use `github.com/fluxcd/helm-controller/api` now

lxm (changes requested):

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,437 @@
...
+
+// ApplicationReconciler is the reconciler of the FluxCD HelmRelease and FluxCD Kustomization
+type ApplicationReconciler struct {
+	client.Client
+	log      logr.Logger
+	recorder record.EventRecorder
+}
+
+// Reconcile sync the Application with underlying FluxCD HelmRelease and FluxCD Kustomization CRD
+func (r *ApplicationReconciler) Reconcile(req ctrl.Request) (result ctrl.Result, err error) {
```
LinuxSuRen, line 51: Considering I have upgraded the controller-runtime, I guess you could rebase with the master branch. 

See https://github.com/kubesphere/ks-devops/pull/788

chengleqi: Sure. will do.

LinuxSuRen: Do you need any kind of help here?

chengleqi: It goes well. Testing now.

Makefile
```diff
@@ -51,7 +51,7 @@ deploy: manifests
 
 # Generate manifests e.g. CRD, RBAC etc.
 manifests: controller-gen
-	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=ks-devops webhook paths="./..." output:crd:artifacts:config=config/crd/bases
+	$(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=ks-devops webhook paths="./pkg/api/..." output:crd:artifacts:config=config/crd/bases
```
LinuxSuRen, line 54: The RBAC annotations exist in both API and controller. You could exclude the `pkg/external`.

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,453 @@
...
+	if at, err = isHelmOrKustomize(app); err != nil {
+		return err
+	}
+	if at == HelmRelease {
+		return r.reconcileHelm(app)
+	}
+	if at == Kustomization {
+		return r.reconcileKustomization(app)
+	}
+	return
```
LinuxSuRen, line 81: Using `switch` statement would be more clear. Such as:
```go
	switch at {
	case HelmRelease:
	case Kustomization:
	default:
		err = fmt.Errorf("should has one and only one FluxApplicationConfig (HelmRelease or Kustomization)")
	}
```

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+func (r *ApplicationReconciler) reconcileHelm(app *v1alpha1.Application) (err error) {
+	ctx := context.Background()
```
LinuxSuRen, line 85: Please use the context from https://github.com/kubesphere/ks-devops/pull/743/files#diff-5da9e5f79765c711202e5d041f78644d98f9502949c1102beebcfbed92c73ad9R51

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,453 @@
...
+
+	var helmChart *sourcev1.HelmChart
+	// 1. get helmChart by searching existed helm template
+	if helmTemplateName := app.Spec.FluxApp.Spec.Config.HelmRelease.Template; helmTemplateName != "" {
```
LinuxSuRen, line 89: `app.Spec.FluxApp.Spec.Config` is a pointer. What happens if it's nil?

controllers/fluxcd/application-controller.go
```diff
@@ -0,0 +1,453 @@
...
+
+func wantSaveHelmTemplate(app *v1alpha1.Application) bool {
+	if v, ok := app.GetLabels()[v1alpha1.SaveTemplateLabelKey]; ok {
+		return v == "true"
+	}
+	return false
+}
+
+func buildTemplateFromApp(app *v1alpha1.Application) *sourcev1.HelmChart {
+	s := app.Spec.FluxApp.Spec.Source.SourceRef.DeepCopy()
```
LinuxSuRen, line 336: `app.Spec.FluxApp.Spec.Source` is a pointer. It would be nil. It's not safe to use those values directly. I know you already checked if app.Spec.FluxApp is nil somewhere. But it's still not safe if other contributors try to use this function without going through all the code lines.

You could consider only passing `app.Spec.FluxApp` directly. Because it's not necessary to know all the structure.

pkg/api/gitops/v1alpha1/application.go
```diff
@@ -17,12 +17,215 @@ limitations under the License.
...
+}
+
+type FluxApplicationDestination struct {
+	KubeConfig *helmv2.KubeConfig `json:"kubeConfig,omitempty"`
+	// TargetNamespace to target when performing operations for the HelmRelease.
+	// Defaults to the namespace of the HelmRelease.
+	TargetNamespace string `json:"targetNamespace,omitempty"`
+}
+
+type FluxApplicationConfig struct {
```
LinuxSuRen, line 47: The comment on all the exported types is necessary. I don't know why the lint does not find it.

pkg/api/gitops/v1alpha1/application.go
```diff
@@ -347,7 +550,6 @@ type SyncStrategyApply struct {
 // If no hook annotation is specified falls back to `kubectl apply`.
 type SyncStrategyHook struct {
 	// Embed SyncStrategyApply type to inherit any `apply` options
-	// +optional
```
LinuxSuRen, line 350: Why?

pkg/api/gitops/v1alpha1/constants.go
```diff
@@ -38,3 +38,18 @@ const ApplicationFinalizerName = "application." + GroupName
 const ArgoCDResourcesFinalizer = "resources-finalizer.argocd.argoproj.io"
 
 const ArtifactRepoLabelKey = GroupName + "/is-artifact-repository"
+
+type Engine string
+
+const (
+	ArgoCD Engine = "argocd"
+	FluxCD Engine = "fluxcd"
+)
```
LinuxSuRen, line 47: Comments are required.

LinuxSuRen (changes requested):

LinuxSuRen:
Please try not to use force action. I need to review all files one by one if you push the code with the force option.

pkg/api/gitops/v1alpha1/application.go
chengleqi: Sorry. It's a miss.

controllers/fluxcd/application-controller.go
chengleqi: Thanks. Really appreciate your advice.

Makefile
chengleqi: You're right. But I have no idea to how exclude the `pkg/external` path in this situation.

LinuxSuRen (commented):
/lgtm

controllers/fluxcd/application-controller.go
```diff
@@ -68,25 +68,28 @@ func isFluxApp(app *v1alpha1.Application) bool {
...
 	var at AppType
-	if at, err = isHelmOrKustomize(app); err != nil {
-		return err
-	}
-	if at == HelmRelease {
-		return r.reconcileHelm(app)
-	}
-	if at == Kustomization {
-		return r.reconcileKustomization(app)
+	at, err = isHelmOrKustomize(app)
```
lxm, line 74: missing error handle

controllers/fluxcd/application-controller.go
```diff
@@ -268,12 +295,14 @@ func (r *ApplicationReconciler) updateKustomization(ctx context.Context, app *v1
 }
 
 func isHelmOrKustomize(app *v1alpha1.Application) (AppType, error) {
-	conf := app.Spec.FluxApp.Spec.Config.DeepCopy()
-	helm, kus := conf.HelmRelease != nil, conf.Kustomization != nil
-	if helm && !kus {
+	if app.Spec.FluxApp.Spec.Config == nil {
```
lxm, line 298: This inspection method has appeared many times, how about change it to a function

lxm (changes requested):

lxm:
lgtm @kubesphere/sig-devops @LinuxSuRen

LinuxSuRen (commented):
/lgtm

PR #799, opened 2022-08-26
Make gitops app kind to be optional

LinuxSuRen:
### What type of PR is this?

/kind bug

It will be a breaking change if the field kind is not optional.
See the following errors:
```
Unprocessable Entity
Application.gitops.kubesphere.io "ks-devops" is invalid: status.kind: Required value
```

/cc @lxm @chengleqi 

Fixes #

```release-note
None
```

### DIFF
--- a/config/crd/bases/gitops.kubesphere.io_applications.yaml
+++ b/config/crd/bases/gitops.kubesphere.io_applications.yaml
@@ -2023,8 +2023,6 @@ spec:
               kind:
                 description: Engine is the backend GitOps Solutions type
                 type: string
-            required:
-            - kind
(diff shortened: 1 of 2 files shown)


chengleqi:
Yes. You're right. It'll make breaking change for now.
lgtm

chengleqi:
And I think this `kind` field maybe not necessary. We can check the `ArgoApplication` and `FluxApplication` whether is not nil to let the corresponding controller to do the reconcile thing.

LinuxSuRen:
We can keep it for now until there's a solid reason to remove it.

PR #813, opened 2022-09-05
Fix nil pointer issue when building helmchart

chengleqi:
### What type of PR is this?

/kind bug

We should provide the nil pointer check and set a default value (10m0s) to the HelmChart's Interval field.

Fixes #

```release-note
None
```

### DIFF
--- a/controllers/fluxcd/application-controller.go
+++ b/controllers/fluxcd/application-controller.go
@@ -33,6 +33,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
+	"time"
 )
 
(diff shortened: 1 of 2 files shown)


controllers/fluxcd/application-controller.go
```diff
@@ -492,6 +493,13 @@ func convertKubeconfig(kubeconfig *helmv2.KubeConfig) *kusv1.KubeConfig {
 	return &kusv1.KubeConfig{SecretRef: kubeconfig.SecretRef}
 }
 
+func convertInterval(interval *metav1.Duration) metav1.Duration {
+	if interval == nil {
+		return metav1.Duration{Duration: 10 * time.Minute}
```
LinuxSuRen, line 498: Please feel free to add unit tests, especially for this line.

LinuxSuRen (changes requested):

controllers/fluxcd/application-controller.go
chengleqi: Thanks, will do

LinuxSuRen (approved):

